PR #96, opened 2015-02-03
Fix typo : changed notofy for notify

natcl:


dceejay:
Hi Nat, sadly before we can accept pull requests we need a signed CLA as per the contribution guidelines :-(....  as it's only a tiny nit in the package.json I'm happy to correct it by hand if you like ?

natcl:
Sure, no worries !

dceejay:
@natcl Thanks for spotting it

PR #116, opened 2015-06-10
added Moni.ai Channel

moni-ai:
Hi guys,

I just wrote a Node-red node for Moni.ai - a virtual assistant for the Internet of Things. With the node you can define how Moni will respond to specific trigger patterns and thereby define your own workflows or applications.


knolleary:
Hi,

we're not generally accepting new nodes into this repository other than the ones we maintain ourselves.

If you follow our [packaging guidelines](http://nodered.org/docs/creating-nodes/packaging.html), you can publish this to npm yourself.

Thanks,
Nick

moni-ai:
Hi Nick,

an IBM representative for Bluemix invited us to contribute on Bluemix. As we just released our developers platform IBM Bluemix with NodeRed would be the perfect fit to allow users to build a speech dialogue/voice interaction. We already published the moni.ai node on npm (https://www.npmjs.com/package/moni.ai) however the procedure for installing this node on the nodered boilerplate on bluemix is too difficult. In order to being able to promote nodered/bluemix we would appreciate having a solution like the twilio node. As I understand therefore it is necessary to have the node in this repository.

best,
Christian

knolleary:
Hi Christian

If you have published the node to npm already, then it does not need to be
in this repository - you can publish it yourself to github and npm. We can
then list your node as a dependency of the Bluemix node-red template so it
will get pulled in when someone deploys a new instance. Otherwise, if it is
our repository we become responsible for maintaining your code.

As per our packaging guidelines, we prefer if nodes use the name prefix
"node-red-contrib-" so it is clearly not a node maintained by us directly.

I'm out of the office the rest of the week, so may be slow to respond.

Nick

moni-ai:
Ok, I updated the npm package. You can now find it under https://www.npmjs.com/package/node-red-contrib-moniai. What do we need to do in order to get it in this node-red template?

knolleary:
Great - we'll take a look, provide any feedback on the node itself. Once
everyone's happy, we will add it to the template and publish it.

As I mentioned, I'm out of the office (currently sat  on a train on our way
on holiday), so won't be able to look at it properly until next week.

My colleague @dceejay can take a look and provide feedback in my absence,
but the final push to Bluemix needs my input.

Nick

dceejay:
Hi,

looks pretty good. Just a few niggles that could be polished (imho).
In general we would like if the labels could also have an icon with them. They are all from the font-awesome fa-icons set - so pick anything suitable.

**Input node** - don't really need the Mon.ai in front of username - as the whole node is Mon.ai...
Trigger pattern ... yes it's long for a label... maybe it could be a whole line above the input box
* Trigger pattern for voice channel
     [             ]
link in info  - ideally add  target="_new" - so the edit page remains open....

**Output node** - just add an icon
(clarification question why "must" the input come from a mon.ai input node ? - most nodes do preserve properties passed in - but not all... is all to easy to strip them in a function... what happens if they are stripped?)

**Ask node** - just add an icon
(ditto question... can the ask node not be triggered by anything other than a mon.ai node ?)

moni-ai:
agree. I updated the npm package. It contains your suggested changes.
Regarding your question: The moni.ai input node works similar to the "http in" node. That means the http request must be terminated by the moni.ai output node (like the "http in" request is terminated by the "http response" node). Currently moni.ai only works in pull mode - that means Moni will trigger the input node and then expects a response via the output node (or the ask node in case further userinput is required).But it is not possible to push a message to Moni.ai. That is because the user is not always available and currently we don't queue messages. Maybe we'll introduce something like that in near future.
One sidenote:
following your instructions on https://developer.ibm.com/answers/questions/180359/node-red-instructions-for-adding-a-new-node.html I was able to build and deploy the application without errors. I can also see that the node is theoretically installed and I see the files in the vm. However like "nksky" I cannot see the new nodes on the gui. That means we were only able to test the nodes locally but not on bluemix.

moni-ai:
So I was able to resolve the "sidenote problem" - the keywords in the package description were lacking "node-red". Now the node displays on bluemix when building the code.

dceejay:
Closing this as now published direct to npm, and working in Bluemix.

PR #119, opened 2015-06-30
added msg.params as an option for twitter out node

sacbhale:
Added support for the "twitter out" node to be able to support additional parameters to twitter. For example twitter needs 'in_reply_to_status_id' field to be set to the value of the original tweet if you want to reply to an existing message. 
it can be set as msg : { params: { in_reply_to_status_id: msg.tweet.id_str })
Other parameters can be found at https://dev.twitter.com/rest/reference/post/statuses/update


dceejay:
Hi,
seems like a sensible idea - but as per our contribution guidelines we need a signed CLA before we can accept pull requests... https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md
So ... given that it's a small fix you can either do the CLA... or let us implement locally :-)

sacbhale:
Too much work to do the CLA. Please implement locally. :)

PR #129, opened 2015-08-15
Allow Pushover notifications with priority -2.

perja12:


dceejay:
Hi,
thanks for that - as per our contribution guidelines we need a completed CLA before we can accept a PR... :-(... but as it's a simple change I'm happy to apply it manually if you don't mind.

perja12:
Thanks, that would be great. I will send the CLA anyway in case I have more changes.

dceejay:
Thanks.

Issue #141, opened 2015-11-02
node-red-node-stomp - Configure the protocolVersion and vhost that are passed to stomp-client.


ozomer:
We are working with CloudAMQP, where these parameters are important: `protocolVersion` should be '1.1' instead of '1.0' and `vhost` has to be a given value.

Being able to configure the `reconnectOpts` parameter should be nice too - it only has two relevant fields: retries and delay (which makes me wonder why they are passed via `reconnectOpts` and not directly...).

The `tls` parameter can have many configurable fields, so I think it should be left-out for now.

ozomer:
The pull-request failed due to some CLA lawyer-stuff issue.
https://github.com/node-red/node-red-nodes/pull/142
Feel free to use my code-changes. I don't request any legal rights for it.
http://www.wtfpl.net

dceejay:
A comment in the stomp-client npm docs says  "Protocol version negotiation is not currently supported and version "1.0" is the only supported version."  and it also says "Ability to support different protocol versions (1.0 or 1.1) - more work needed" - so did you try 1.1 ? does it actually work ?

ozomer:
I am using protocol 1.1 in order to connect to the stomp plugin of CloudAMQP. In fact, if you specify protocol 1.0 the connection will fail.

I suppose 0 retries (i.e. no retries) is a sensible value. I don't know if a 100ms repeat-delay is sensible, but notice that in the stomp-client implementation the delay is increasing: the first retry will be after 100ms, the second retry will be 200ms after the first retry, the third retry will be 300ms after the second retry etc.

PR #142, opened 2015-11-02
Using stomp-client version 0.8.0

ozomer:
Following: https://github.com/node-red/node-red-nodes/issues/141
Allowing to configure additional parameters to the stomp-client: protocol-version, vhost, reconnect-retries, reconnect-delay.


knolleary:
Hi, thanks for this. As you've indicated you don't plan to complete the CLA, I have to close this PR.

We'll look at the original issue in due course and add the necessary changes.

ozomer:
OK. One more minor issue - I noticed that the default port of node-red-node-stomp is 61618, but the default port of stomp-client is actually 61613. I didn't change it, but maybe you should do so if you merge the code anyway.

PR #151, opened 2015-12-07
fix for OAuth on older browsers (e.g. Epiphany on Pi)

andypiper:
This tiny change fixes an issue with the Twitter node on devices like the Raspberry Pi.

Previously, the authentication step went to twitter.com, which redirected to mobile.twitter.com on older Webkit versions, where the oauth pin step is not supported, leaving the developer unable to add Twitter credentials.

With the change, it now sends the user directly to api.twitter.com which will work in browsers like the Epiphany version shipped in Raspbian.

Confirmed to work as a fix on a Pi Zero with latest Raspbian.


knolleary:
Thanks! :+1:

PR #158, opened 2015-12-23
Add support for pushover sound selection.

Shirk:
Pushover supports a list of [notification sounds](https://pushover.net/api#sounds) to select from when pushing a notification.

This PR extends `PushoverNode` to honor the optional `msg.sound` field for this purpose.
If `msg.sound` is specified and contains a string it will be passed as the sound parameter.
If it's not set or not a valid string the same behavior as before is used.

### DIFF
--- a/social/pushover/57-pushover.html
+++ b/social/pushover/57-pushover.html
@@ -47,8 +47,8 @@
 
 <script type="text/x-red" data-help-name="pushover">
     <p>Uses Pushover to push the <b>msg.payload</b> to a device that has the Pushover app installed.</p>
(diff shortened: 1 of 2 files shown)


bigmonkeyboy:
Thanks - have seen CLA discussion on mailing list.

As there is a limited set... - do you also want to add a select box to the config dialog to allow user to set manually also ? default should be - nothing selected... and if nothing selected then msg.sound can override.

Shirk:
Good idea, I'll update the PR :smile:

dceejay:
node-red-gitbot: check cla status

Shirk:
node-red-gitbot: check cla status

knolleary:
let's try that again - node-red-gitbot, do your thing.

PR #168, opened 2016-01-18
Add SensorTag connection status notifications.

jthomas:
Use Node-RED status messages to indicate the connection status of the TI Sensor Tag nodes used in flows.


hardillb:
Looks good to me

PR #183, opened 2016-03-01
Added support for sending and receiving strings

jpmanson:
Added support for sending and receiving strings. The modifications uses the ability to send SysEx messages supported in arduino-firmata.js library. The string is encoded as a SysEx message of type 0x71 as specified in Firmata protocol.


dceejay:
Hi,

Line 54 of the html file - has required:true to make sure that a pin is specified in order for the node to be configured... Obviously for String this is not required...  this needs to be handled in some way...

jpmanson:
One solution could be to disable the Ok button if Pin is NaN and Type is Digital/Analogue/Servo. Or remove the required clause for Pin field...

dceejay:
Generally we don't disable the OK button, maybe do the just not required for now... 
also would be good if you can squash the commits down into one.. thanks

jpmanson:
I will generate a new PR with only one commit

PR #184, opened 2016-03-02
Added support for sending and receiving strings

jpmanson:
- Now the changes are in only one commit
- The Pin field has been set to not required


dceejay:
Great thanks !

PR #185, opened 2016-03-04
SQL error with HTTP request

paulRenault:
In the case the SQL node is in a flow with http in and out, the flow end
in case of error but the http client is waiting for an answer. So we
must not endded the flow but send back the error.

### DIFF
--- a/storage/mysql/68-mysql.js
+++ b/storage/mysql/68-mysql.js
@@ -103,8 +103,11 @@ module.exports = function(RED) {
                     //console.log("query:",msg.topic);
                     var bind = Array.isArray(msg.payload) ? msg.payload : [];
                     node.mydbConfig.connection.query(msg.topic, bind, function(err, rows) {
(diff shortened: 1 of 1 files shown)


knolleary:
Hi,

you'll see in the code you removed there was a call to `node.error(err,msg);`.

That error can be caught by a `Catch` node on the same tab as the sql node, and the message passed to that node. That allows you to create an error-handling flow that will respond to the HTTP request appropriately.

No changes are needed to the node itself.

Nick

paulRenault:
Hi,

The HTTP in and HTTP out need to be in the same flow to share the headers ?
Am I wrong ?

Paul

knolleary:
The key thing is the message object that gets passed through. The Catch node is given the message from the sql node so it contains everything needed to respond to the originating http request.

paulRenault:
OK

Thanks

Issue #187, opened 2016-03-13
node-red-node-physical-web: Period is not the actual Advertising period.


natcl:
After some digging into eddytsone-beacon and bleno, I found out the actual advertising interval is fixed at 100ms in bleno.  The Period parameter that node-red-node-physical-web actually changes is the ratio between regular messages and TLM (telemetry) packets sent.  I'm not sure this ratio is very useful to expose in the node and might confuse people...

If one wants to change the actual advertising interval, you must set an environment variable, example:
```
sudo BLENO_ADVERTISING_INTERVAL=500 node <your file>.js
```

It also seems that on OSX the rate is always set to 100ms because of OS limitation.

dceejay:
closed by PR above https://github.com/node-red/node-red-nodes/pull/188/commits/6d41d33e806cdfd24040afe71722164bd44db118

PR #188, opened 2016-03-14
Added uid mode to node-red-node-physical

natcl:
Added uid mode to node-red-node-physical and exposed the count property to make the period setting more relevant.  Updated the docs.

Thanks !


natcl:
Note: this closes #187

dceejay:
Thanks @natcl

Issue #192, opened 2016-03-20
unexpected result with rbe node


greyltc:
I'm using v0.13.3 and I have a rbe (report by exception) node configured as `block if value changes by more than` `500` with the Start Value set to be `9999`.

If I send a msg.payload of 14 twice consecutively to my rbe node, I expect the first message to be dropped (since abs(9999 - 14)>500) and the second message to make it through (since abs(14-14)<500), but instead they both get dropped.

Is this a bug or is this intended behavior?

dceejay:
same reply as before...

good question... the current behaviour is that if a value is outside the desired range it is ignored (and thus doesn't count towards the next comparison)... so the second comparison in your example is still abs(9999 -14) > 500 - so also gets ignored...

PR #193, opened 2016-03-20
rbe: always remember previous value

greyltc:
This change addresses https://github.com/node-red/node-red-nodes/issues/192

Previously the rbe node was only "remembering" the last numerical value it saw if it forwarded the message. This change means it always remembers the previous value, even if it blocks the message.

### DIFF
--- a/function/rbe/rbe.js
+++ b/function/rbe/rbe.js
@@ -59,16 +59,15 @@ module.exports = function(RED) {
                         if (!node.previous.hasOwnProperty(t)) { node.previous[t] = n - node.gap; }
                         if (Math.abs(n - node.previous[t]) >= node.gap) {
                             if (this.func === "deadband") {
-                                node.previous[t] = n;
                                 node.send(msg);
                             }
                         }
                         else {
                             if (this.func === "narrowband") {
-                                node.previous[t] = n;
                                 node.send(msg);
                             }
                         }
+                        node.previous[t] = n;
                     }
                     else {
                         node.warn(RED._("rbe.warn.nonumber"));


greyltc:
My completed CLA is attached [here](https://github.com/node-red/node-red-nodes/files/181304/node-red-cla-individual.pdf), also emailed to Nick O'Leary.

dceejay:
Why is this behaviour desired ? This will mean that sending two "bad" values in a row will force through the second "bad" value...

greyltc:
I was hoping this node would act as a simple filter for single outliers in narrowband mode without totally blocking step changes.

> Why is this behaviour desired ? This will mean that sending two "bad" values in a row will force through the second "bad" value...

@dceejay, In my application, if my sensor produces two consecutive values that agree with each other, then that's enough evidence to say that data is probably not actually bad. Comparison to the previous input rather than the previous output (the change made in this PR) blocks single outliers without causing the node to block good data that is gathered after a sudden change. This is critical for my application, I also thought this fix brought the node in line with its intended function.

I guess I was fooled by the description in the UI. Here's what I see in the Info tab for the node:
> Report by Exception node - only passes on data if it has changed.
> 
> The node can either block until the msg.payload is different to the previous one - rbe mode. This works on numbers, strings, and simple objects.
> 
> Or it can block until the value changes by a specified amount - deadband mode.
> 
> In deadband mode the incoming payload must contain a parseable number and is output only if greater than + or - the band gap away from the previous output.
> 
> Deadband also supports % - only sends if the input differs by more than x% of the original value.
> 
> It can also ignore outlier values - narrowband mode.
> 
> In narrowband mode the incoming payload is blocked if it is more than + or - the band gap away from the previous value. Useful for ignoring outliers from a faulty sensor for example.

I can't imagine anyone who reads the narrow band mode section there would ever know that the code actually compares between the gap specified and the difference between the current input and whatever the last output value was. I did miss the final word "output" in the deadband section above, where it's made clear what's going on there, skipped that section since I wasn't using that mode. The wording should probably be made more clear if you intend to leave the code like this.

Here are a few scenarios when the node's present narrowband mode design (the only data that's allowed through is data that's similar to the last "good" data) could cause trouble:

- The node becomes latched in an undesirable state (blocking forever) if by chance the first data point is the errant data that it's meant to be filtering out.
- The node can't recover from legitimate step changes. Maybe the user is measuring a value corresponding to the wear level of a machine and then it gets repaired. Here the measured wear metric slowly drifts up, and the node continuously re-trains its self on what "good" data looks like. Then when the machine is fixed, the wear metric quickly falls back to normal. The node unexpectedly blocks because of its skewed notion of what "good" data looks like.
- The node's behavior is unexpectedly sensitive to initial conditions. Maybe a user has set up a flow which collects pressure data for a pump that runs intermittently. The pump happened to be running today when she clicked the Deploy button in the UI so she only recorded high pressure values corresponding to times when the pump was on because the node was trained to think only high pressure measurements are good. Yesterday she hit the deploy button when the pump happened to be off and so her pressure data only flowed during times when the pump was off. Troubleshooting these sorts of sensitivities to initial conditions in a complicated system is brutal.

philippe-gregoire:
Would it make sense to have this behavior as an option?

e.g. 'Compare using last output' vs 'Compare using last input' ?

Philippe Gregoire - IBM Innovation Center Nice - ISVs & Business 
Development
Phone: IBM Office +33-4-9296.8683 Mobile: +33-6-8482-8091

greyltc:
Yep, I think so!

dceejay:
Yes - probably :-) 
Currently you can leave the start value blank - and on start it waits to see what the first value is and uses that as the set point to work from. 
Waiting for two readings... what happens when someone wants to wait for three readings, or 4  ?
What about accepting a msg.reset (or msg.start) value to set/reset it ?

(and yes we can accept PR for docs/info also :+1: )

dceejay:
I think Philippe's suggestion encompasses both modes elegantly. @greyltc do you want to code that up ?  or let me do it ? (I don't mind either way, but if you do please use a select box for the two options).

greyltc:
This 3 line PR was the first time I've ever touched JavaScript; I'd be happy if you'd were willing to do it.

dceejay:
:-) - no probs

PR #200, opened 2016-04-12
Added subtree and walker nodes.

mikakaraila:
More efficient way to read OID from OID node underlying leafs.
Walker to print out existing OIDs and descriptions.


knolleary:
node-red-gitbot: please update cla check

PR #203, opened 2016-05-02
Fix repository URL for smooth

Ltrlg:


dceejay:
Hi, I'm afraid we require a signed CLA before we can accept a PR as per the overall project contribution guidelines. However as as it's a simple one liner I'm very happy to do it manually if that's ok with you ?

Ltrlg:
I saw it only after I made it. I will consider signing the CLA if I have bigger PRs, but you can do this one manually.

dceejay:
great. thanks for spotting it.

PR #206, opened 2016-05-18
Physical-web: Added status (found, updated and lost) as msg.topic

natcl:
Physical-web: Added status (found, updated and lost) as msg.topic
Added the ability to filter duplicates.

### DIFF
--- a/hardware/physical-web/package.json
+++ b/hardware/physical-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "node-red-node-physical-web",
-  "version": "0.0.12",
+  "version": "0.0.13",
   "description": "Node-RED nodes to interact with the Physical Web",
   "main": "physical-web.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


dceejay:
Nat, you have removed the existing topic field so that it can no longer be specified.... this will break anyone who happens to be using it today.

natcl:
Hmm, how can we manage this ? Setting a static topic on this node is not very useful...

dceejay:
you can add another property msg.status ? msg.action ?

natcl:
yeah that's what I did in the beginning but didn't find it elegant since the topic isn't really useful...
Could we perhaps re-add the static topic and if it's define it will be sent and if it's blank the status will be sent instead ?

dceejay:
yes - that is actually a great answer... make the topic placeholder (optional) and add some info to document the behaviour.

natcl:
Cool will do, thanks !

natcl:
Fixed !

PR #208, opened 2016-05-30
Fixed "update" operation

jjfattz:
"Update" operation was converting string payloads into an object by adding the "payload" key.  I believe we just want the value of the payload object.  So instead, the JSON.parse function was used to convert the payload string into an object without adding a "payload" key.

### DIFF
--- a/storage/mongodb/66-mongodb.js
+++ b/storage/mongodb/66-mongodb.js
@@ -124,7 +124,7 @@ module.exports = function(RED) {
                             }
                         } else if (node.operation === "update") {
                             if (typeof msg.payload !== "object") {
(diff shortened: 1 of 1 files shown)


dceejay:
The update option can also do an upsert. In either case it requires a full document and not just a single value. Also just doing JSON.parse is very fragile without any try catch and what do you do if the string is not parseable. (eg if some passes in a value "banana".  

We made the conscious decision (also on the insert and save options) that if someone "just" passes in a value to wrap it into a document that could then be saved. 

Also changing the existing behaviour will now break some people without sufficient deprecation. 
Overall I can't see a reason just to do this as-is.

PR #211, opened 2016-06-07
Collects every unseen email

bkems:
This version collects all unread emails and sends them through individually until they’re all processed. It does up to 10 in every iteration.


dceejay:
Hi, thanks for these, The current master does handle all unread messages.
Also the beautifications and tabs are not required as they basically change every line in the files making diff somewhat irrelevant. - There is a .jscsrc file in the root of the node-red-nodes project with our preferred style.
However I see your other PR - which does indeed fix things we need so I assume this is a way to get to that ?
Finally we also need a signed CLA as per the contribution guidelines before we can accept a PR - but as this PR may be redundant that may not be necessary.

dceejay:
Hi, as noted on your other PR we need a signed CLA as per the contribution guidelines before we can accept a PR - however as this is a small change I can apply manually if you like . ?

bkems:
Hi, that's fine by me. :-)

Sent from Outlook Mobile<https://aka.ms/qtex0l>




On Tue, Jun 7, 2016 at 5:39 PM +1000, "Dave Conway-Jones" <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi, as noted on your other PR we need a signed CLA as per the contribution guidelines before we can accept a PR - however as this is a small change I can apply manually if you like . ?

-
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub<https://github.com/node-red/node-red-nodes/pull/211#issuecomment-224203703>, or mute the thread<https://github.com/notifications/unsubscribe/AHfepy_Lz27tfw2SQQi9IvR6Wnb-AmXrks5qJSALgaJpZM4IveQN>.

This e-mail is confidential and may be read, copied and used only by the intended recipient. If you have received it in error, please contact the sender immediately by return e-mail. Please then delete the e-mail and do not disclose its contents to any other person.

dceejay:
Hi, great - will make it so. (And as it happens there's some other nonsense I'm trying to clear up in there anyway - so merging would have been less than fun  ;-)

PR #212, opened 2016-06-07
Fixed 2 bugs

bkems:
node.status was not being reset when no email found to be processed, and imap connection not reset before the return, which left open imap connections.


dceejay:
Hi - unfortunately we need a signed CLA before we can accept a PR as per our contribution guidelines. However as it's a small change I'm happy to update manually if you like. ?

PR #215, opened 2016-06-27
Removing Hardcoding for Inbox Folder

aamol:
Currently Inbox is hardcoded which stops from using any other folder from being used of email mail box. 

Fixing the same


dceejay:
Hi - yes thanks - good spot. Sadly we need a signed CLA before we can accept a PR - but I'm apply to apply this simple fix by hand if that's OK with you ?

aamol:
Yes. Please thats fine. I am only concerned with Fix :smile:

aamol:
By When will I get that fix in build

PR #223, opened 2016-08-02
Fixed #222 - RFC error with invalid FROM address with userid conf

### LINKED ISSUE #222: node-red-node-email RFC issue blocking emails
Email sent via Sendgrid SMTP to address hosted by Google is blocked and never sent. The error states the email is incorrectly formed according to RFC specifications. Emails send to other addresses but for some reason Google blocking them?

I suspect since Sendgrid sets the reply address to their bounce server, the RFC 5322 From address fails with our SPF or other spam protection. We might need ability to set an additional field when sending emails, the RFC 5321 MailFrom field, to prevent this error. Does anyone have a workaround or face this issue?

screen shot 2016-08-02 at 9 13 09 am

~~~
Our system has detected that this message is not RFC 5322 compliant. To reduce the amount of spam sent to Gmail, this message has been blocked. Please review RFC 5322 specifications for more information. g13si16656837ote.5 - gsmtp
~~~

mikesparr:
Added ability for user-defined `from` address in the `msg.payload`.  If `from` not in `msg.payload` then will default to `userid` but allowing override in case of #222 where `userid` is not a fully-qualified email address but a 3rd-party provider username (i.e. Sendgrid).

## Example usage
~~~
msg.payload = {
    ...
    "payload": {
        "to": msg.newRecord.recipients.map(function(email){
            return {"email": [{"value": email}]};
        }), 
        "from": "No-Reply Address <no-reply@domain.com>",
        "title": msg.newRecord.title,
        "body": msg.newRecord.body
    }
};
~~~

### DIFF
--- a/social/email/61-email.html
+++ b/social/email/61-email.html
@@ -70,8 +70,8 @@
 
 <script type="text/x-red" data-help-name="e-mail">
     <p>Sends the <code>msg.payload</code> as an email, with a subject of <code>msg.topic</code>.</p>
-	<p>You may optionally set <code>msg.from</code> in the payload which will override the <code>node.userid</code> 
-	default value.</p>
+    <p>You may optionally set <code>msg.from</code> in the payload which will override the <code>node.userid</code> 
+    default value.</p>
     <p>The default message recipient can be configured in the node, if it is left
     blank it should be set using the <code>msg.to</code> property of the incoming message. If left blank
     you can also specify <code>msg.cc</code> and/or <code>msg.bcc</code> properties.</p>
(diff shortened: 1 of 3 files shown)


dceejay:
Hi Mike, thanks for this.
Sadly we need a CLA as per our contribution guidelines - but as it's a simple change I'm happy to apply manually if you like.... - I'll also need to add some words to the info panel to explain this extra option as normally we don't like msg properties to override things set in the edit dialogue

mikesparr:
I did complete one and sent to Nick O'Leary in UK IBM office. I assume he'll process tomorrow (GMT). Your choice on how you implement, but I think it's important that users have the option to declare a `from` value because SMTP userid will not always be a fully-qualified email address.

mikesparr:
Also, I think rather than the node setup (HTML) as I saw you had commented out a drop-down list of providers, it's better this way to allow the app to declare the value (more dynamic) rather than just hard-coded at node config level.

knolleary:
node-red-gitbot: please recheck cla status

dceejay:
ah great. Yes agree needs to be done. And yes agree that in this case it makes sense to allow the msg to override - (as normally we don't allow override if you can leave the edit value blank - but as this is the userid you can't...) . So I'm happy to accept this PR... but would you like to add a comment to the help info in the html first ?

mikesparr:
@dceejay It should be good now with both hints in README and the HTML file. Thanks for considering and replying so quickly.

dceejay:
Thanks Mike

PR #230, opened 2016-09-05
adding STT support for email node

wajnberg:
Signed-off-by: Moshe Wajnberg <wajnberg@il.ibm.com>


wajnberg:
Hello @knolleary 

Could you please review this very simple change ?

Thank you

dceejay:
We need to be extra careful with non-core nodes, as they could be installed on versions of Node-RED without the required patches... Will this break them ?

PR #235, opened 2016-09-21
Twitter node: Add support for search terms passed in through msg.payload

smchamberlin:


dceejay:
Our standard pattern at the moment is only to allow input to be over-ridden if the field in the edit dialogue is blank... so potentially the input pin could be enabled/disabled by that also - though we don't actually do that anywhere else.

(Mainly I guess because the input may have other things on it anyway - but in this case if that is the only thing that can be passed in... )

PR #240, opened 2016-10-17
Added support for 'url' and 'url_title' for Pushover

eisbehr-:
Added support for `url` and `url_title` properties of the Pushover API message object. These where missing in the current version. There is actually no other way to use these features with pushover. It's just two other optional properties in the message object. I've used the same coding-style as for the other properties in the original file.

Further information about these properties can be found in the docs:
https://pushover.net/api

### DIFF
--- a/social/pushover/57-pushover.js
+++ b/social/pushover/57-pushover.js
@@ -48,6 +48,8 @@ module.exports = function(RED) {
             var pri = this.priority || msg.priority || 0;
             var dev = this.device || msg.device;
             var sound = this.sound || msg.sound || null;
+            var url = this.url || msg.url || null;
+            var url_title = this.url_title || msg.url_title || null;
             if (isNaN(pri)) {pri=0;}
             if (pri > 2) {pri = 2;}
             if (pri < -2) {pri = -2;}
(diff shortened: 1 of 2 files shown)


hardillb:
@eisbehr- Nick is out a conference for a couple of days so I can't check. Have you submitted a CLA? If not please read the [CONTRIBUTING.md](https://github.com/node-red/node-red-nodes/blob/master/CONTRIBUTING.md) for details as we can't accept any pull requests without one.

eisbehr-:
Hello @hardillb and thanks for your response! No, I haven't sign a CLA. And I will not do it too. Because that is not what "social coding" means to me, if I have to give you my data before. So it's up to you ...

Feel free to decline my pull request or implement it on your own. I don't mind. But I think you should support all features of an API and not only half the properties (and the `timestamp` property is still missing and should be implemented too). Especially when it is so easy as in this case. :wink:

dceejay:
Hi,

thanks for the PR - as it's simple change we may well add manually seeing as you don't mind.
The (possibly) good news is that we are transferring governance to the newly formed JS foundation - https://js.foundation/ 

This will have a slightly different contribution process - but we are not quite sure on all the details yet (most probably self signed DCO but... )... We will update the site as soon as we know and confirm and maybe that will be more to your liking :-)

eisbehr-:
Thank you very much!

PR #255, opened 2016-11-26
Update 100-heatmiser.js

GadgetUK:
Change to underlying nodejs module for addressing Neo and Wifi units


dceejay:
Thanks for that - have bumped version on npm

PR #257, opened 2016-12-10
Feature/mysql node query input

ryansh100:
When I was using this node, I found it a little strange and cumbersome to keep setting the topic to a sql query. I also felt that escaping input variables could be done a little more verbosely.


knolleary:
Hi - please have a read of our contribution guidelines https://github.com/node-red/node-red-nodes/blob/master/CONTRIBUTING.md

We prefer to discuss proposed changes before PRs are opened. There may be merit to the issues being addressed, but the details of implementation need to be discussed.

Can you provide a succinct summary of the changes to the externals of the node rather than expect us to study your changes to figure out everything you've changed? For example, in this PR, you've changed lots of things you don't mention in your one line description in this PR - the node colour/icon etc.

ryansh100:
Sorry for being so slow to reply. 

One of the main concerns that I had is that the mysql node is that it makes the expectation that the sql query is in the topic, which requires unnecessary nodes to shift around properties to preserve the original message topic, when mysql just happens to be a step in the flow. 

My second concern was that escaping properties didn't seem to be a default behavior; however possible as a second parameter of ordered array.

My primary change was to add a SQL query editor that stores the query directly in the node. I also implemented a template literal escaping of the sql query to allow for flexible ordering of data to query inputs. The color and icon, are subtle UI improvements that helped me distinguish between the original and modified during development.

I did maintain backwards compatibility to avoid any breaking changes.

dceejay:
Hi - the templating you are adding uses {... we already have mustache which uses {{ - and angular in the dashboard that uses {{ - so... would it make sense to "standardise" to {{ - or indeed use mustache if necessary ?

ryansh100:
Template literal is part of ES2015 standard so it presumably is here to stay; however, I switched to moustache style since NodeRED has a lot of other moustache touch points. 

I have also setup testing around the query creation.

ryansh100:
Should I abandon this PR?

hardillb:
We can't do anything until everybody signs the CLA

ryansh100:
Ohhhh ok, I don't think I had an email at all on my early commits. Is it possible to amend those commits? I can also re-do it.

hardillb:
Given that the changes also clash with the head of the codebase it may be easier to scrap this PR and start again to ensure the updates apply cleanly and all the updates are owned by one userid/email address that has signed the CLA

Issue #263, opened 2016-12-20
node-red-node-daemon: output object handling


ctr49:
I have a daemon that produces json data. This can be handy as functions could access objects in the nested json like "msg.payload.object". But unfortunately, regardless of the input handling method (string, lines) my input json object is treated as string and quoted as result.

Am I missing something or is this the intended behaviour?

I created a workaround, not sure if that is acceptable though.

dceejay:
does the json node not convert it back ok ?

knolleary:
see also my comment on the associated pr #264

ctr49:
I haven't tried it. Per my understanding the json node converts between json and object (and vice-versa). The current behaviour is to put a json as string value (with escaped quotation marks) rather than a nested json:
input:
{ "value": "001" }
current behavior (valid json, but value cannot be accessed directly using msg.payload.value):
{ "payload": "{ \"value\": \"001\" }" }
expected behavior (and actual behavior with my patch):
{ "payload": { "value": "001" } }

knolleary:
Try it. It works :)

ctr49:
But why should "daemon" mangle the input in the first place to have a another node clean things up?

knolleary:
It doesn't mangle it. It puts the output received from the process into the 'payload' property of the message it sends. That is entirely consistent with how, by convention, every node operates.

dceejay:
what gets received inside the daemon node is a stream of bytes from the stdout pipe of the child process. There is no concept of an object at that point. As Nick pointed out in  #264 , that could be an option - but why force every node to have that option when the JSON node already does what is needed.

ctr49:
I agree that the approach to change the default behaviour might not be the best. I was looking at it from the perspective of "best effort" handling. If the output of the daemon is already a (valid) object, then why bother and modify it. But this would change the behaviour for existing users, so agreed, not a good idea.

However I still don't understand the concept why perfectly valid output (json object or javascript object) needs to be converted to a string (escaping the structure out of the object) just to use another node to fix it again.

knolleary:
@ctr49: see the comment from @dceejay 

What you describe as 'perfectly valid output' is just a stream of raw bytes coming from the external process. That is what we are given to pass on. If we did nothing, you'd just get a raw buffer object containing those bytes. As it is, we provide the convenience to convert those raw bytes into a string.

The node is not converting a valid Object into a String. It is converting raw bytes to a string, which you may then be able to parse further into a JavaScript Object.

[...]

PR #264, opened 2016-12-20
pass-through objects

ctr49:
workaround for #263


knolleary:
The node already lets the user explicitly state whether the output should be a Buffer or attempt to parse as a String.

Adding a mode where it automatically attempts to parse as JSON based on some other criteria is not consistent. It assumes the only time you may get JSON out is if you sent JSON in.

The current 'workaround' is to simply add a `JSON` node after this node as you know, in your application, that you're getting JSON back from the process.

If we wanted to change anything, and I'm not yet convinced we do, it would be to add an additional option to the list of output types to explicitly request the node to parse the output as JSON.

dceejay:
Closing as wontfix

PR #271, opened 2017-01-28
remove firmata/node_modules path in require('serialport')

BartNetJS:
serialport is not found when installing the npm module local (instead of 'npm install node-red-node-arduino -g')
require will look by default in the node_modules folder


dceejay:
Thanks - was there for historical reasons. No reason for that now.

Issue #253, opened 2016-11-22
Trying to authorize twitter just gives "Unauthorized"


Zordid:
I have no clue what this means or why this happens.

I run Node-RED on an Azure Ubuntu VM with a public IP address but no domain.
I switched on https on my Node-RED and enabled admin authentication, but that's all.

Click on the button "Click here to authenticate with Twitter" immediately brings a new tab with just the word "Unauthorized".

Any clue?

tjeerdytsma:
I have the same problem. Also Ubuntu, enabled login of Node-RED and are on https.
![screen shot 2016-12-12 at 10 50 28](https://cloud.githubusercontent.com/assets/6126045/21094791/d41b8a5c-c058-11e6-8b6c-3aa091592eaa.png)

DaviGuanabara:
Same problem :(
I tried to do this using Google Chrome with Android and El Capitan, and Safari with IOS and El Capitan.

I think that its a bluemix settings problem.

knolleary:
I think this may actually be related to us using our own twitter app credentials in the node, rather than require the user to register their own app credentials. I suspect the twitter api is rate limiting the sheer number of twitter connections coming from the Bluemix range using the node-red app credentials.

The long term solution is to deprecate this node and move to a model where the user has to register their own app and use its credentials - as we do with pretty much all other nodes that access 3rd party services.

tjeerdytsma:
I think if you are using your own credentials that can be indeed the problem. It could explain why it works at my own computer at localhost but not at my server. Lot of API's require to whitelist an IP address first and make an exeption for localhost. Someone knows if this is true for Twitter too, or if they changed this recently?

andypiper:
You don't need to be whitelisted, but yes, too many logins from a single IP range and App ID may cause an issue of this kind, as it would be a potential indicator of unusual behaviour to our antispam systems.

jturner-22:
We're also seeing this issue, and I think I know at least one cause. I believe it was introduced in the 0.1.8 version of the Twitter node, and that it only affects Node-RED installs using authentication.

In [7917ca8d], an authorization check was added to several nodes, including Twitter. A call to     RED.auth.needsPermission() was added, which is actually used in quite a few other places and works fine.

Normally, calls are made with $.ajax() (or $.getJSON(), etc.), which in the [editor/js/settings.js](https://github.com/node-red/node-red/blob/master/editor/js/settings.js) file is configured to add an "Authorization" header with a Bearer token.

[...]

PR #278, opened 2017-02-18
Add user's auth token when authenticating Twitter (#253)

jturner-22:
This PR addresses issue #253, or at least one instance of it.

When initiating the Twitter authentication process, I've added the Node-RED user's access token as a query parameter. This had to be added to both the initial request and the callback URL. A query parameter named "access_token" is supported by the [passport-http-bearer](https://www.npmjs.com/package/passport-bearer-strategy) module used by Node-RED. This then allows the RED.auth.needsPermission() check to pass on both of the Twitter node's server endpoints.

The only other ways to get through the "passport-http-bearer" check is with the user's bearer token in the HTTP header or the body. I don't think either of these would work with the approach used by the Twitter node.

However, I'm uncomfortable that the Node-RED user's bearer token is being passed in the callback URL. That's going to Twitter and back. If that is indeed unacceptable, we can close this PR.

Maybe this needs a cookie-based approach or Twitter's [PIN-based](https://dev.twitter.com/oauth/pin-based) option, both of which would be a big change.

Maybe a compromise would be to remove the authentication check on the callback endpoint.

### DIFF
--- a/social/twitter/27-twitter.html
+++ b/social/twitter/27-twitter.html
@@ -32,9 +32,27 @@
                 if (pathname.slice(-1) != "/") {
                     pathname += "/";
                 }
+
+                // If the Node-RED user is logged in, add their authentication token as a
+                // query parameter to both the button's link and the callback.
+                var authTokens = RED.settings.get('auth-tokens');
+                var accessToken = authTokens && authTokens.access_token;
+
                 var callback = encodeURIComponent(location.protocol+"//"+location.hostname+":"+location.port+pathname+"twitter-credentials/"+twitterConfigNodeId+"/auth/callback");
+
+
+                if (accessToken) {
+                    callback += encodeURIComponent('?access_token=' + accessToken);
+                } 
+
+                var link = 'twitter-credentials/'+twitterConfigNodeId+'/auth?callback='+callback;
+
+                if (accessToken) {
+                    link += '&access_token=' + accessToken;
+                }
+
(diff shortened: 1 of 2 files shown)


knolleary:
Agreed with your comment that we shouldn't be passing the access token on the request.

I think removing the access control on that specific route may be the best way to proceed.

Am on vacation this week - if you want to propose an alternative pr then I'll take a look when I can.

jturner-22:
I reworked the fix a bit, and like it more than the first. Now neither of the requests has the user's authorization token in a query parameter.

 1. __First Request:__ On the original request at '/twitter-credentials/:id/auth' that starts the authentication process, I removed the authentication token from the query parameter.

    Instead, I replaced the "Click here to authenticate with Twitter" anchor's href link with a $.ajax() call. The server now sends a 200 with the Twitter URL that needs to be opened, rather a 302 redirect. The client then opens a new window at that URL.

    Basically, I'm doing my own redirect into a new window, rather than deal with the underlying XMLHttpRequest trying to handle the redirect.

2. __Callback Request:__ As suggested, I removed the authorization check on the callback which is called by Twitter's authorization page.

    This still doesn't seem ideal, but it seems to me that anything else will require much larger changes and is well beyond this immediate fix.

knolleary:
On review, I think it is sufficient to remove the `needsPermission` middleware from both auth routes. I don't immediately see a way that can be abused in a meaningful way.

Have pushed that fix and published a new version of the node to npm.

PR #295, opened 2017-04-02
Big WeMo node update

hardillb:
Adds:

 - New Lookup node
 - Make light groups work
 - Fix dimming support for lights
 - Include capability name in event node output for lighting changes


hardillb:
@dceejay checked the ssdp node, all good to merge an push 0.1.11

PR #297, opened 2017-04-04
Mysql node update for holding query directly.

ryansh100:
Update the mysql node to have a direct query input in the node configuration to free up the msg.topic.

I had a previous pull request out there that was discussed, but didn't pass the CLA because I had some commits without email address set.

This feature attempts to solve the lack of a direct input of SQL query into node instead of over-writing the topic.

Mutating the topic in the middle of a flow seemed very strange to me when using the mysql node.


ryansh100:
I think your coverage tool is not including files with no coverage at all.

JustinAiken:
👍!  I'm fighting side effects from overwriting the `topic` right now

ryansh100:
Justin, would you find value in trying to get this actually merged in?

PR #298, opened 2017-04-04
fix dweet.io href

lukewendling:


dceejay:
Hi,

are you happy to accept the CLA so we can accept the PR ? or I'm happy to close this and apply manually. Up to you.
Thanks for spotting it.

PR #303, opened 2017-04-19
Make msg.limit and msg.skip optional as per the README

hardillb:
Currently not passing in a `msg.limit` or `msg.skip` leads an error rather than an unbound result set.

Added checks for undefined and set the limit and skip to 0 if found.


dceejay:
pushed as v0.0.12

PR #309, opened 2017-05-07
Pisrf performance fix

bartbutenaers:
Hi Dave, Nick,

I had connected a HC-SR04 ultrasone sensor to my Raspberry Pi 3 ( pin 7 = trigger / pin 11 = echo ) and it works like a charm.  Even when my sensor is unpowered but connected (to the GPIO pins), performance still looks good.

However when I enter (in the properties of my node-red-pisrf node) some pin numbers that are unconnected ( pin 16 = trigger / pin 18 = echo ), one of my 4 cores goes up to 100% (measured with my [cpu](https://www.npmjs.com/package/node-red-contrib-cpu) node):

image

Seems the Python script (spawned by the pisrf node) has a time-out check when waiting for a rising edge, but there is no timeout for a falling edge.   When I add the timeout, everything looks to be fine again:

image

Note: I have used a time-out of 5 seconds.  However since the range of these kind of sensors is only about 4 meters, a timeout of 25 or 30 milliseconds should be sufficient.  When I change it to 0.2 (like in the existing check) it still seems to be working fine.  So please adjust the number to whatever you like.

Kind regards,
Bart Butenaers

### DIFF
--- a/hardware/PiSrf/nrsrf.py
+++ b/hardware/PiSrf/nrsrf.py
@@ -33,6 +33,12 @@ def Measure():
             Main()
     while GPIO.input(ECHO)==1:
         stop = time.time()
+        Dif = time.time() - realstart
+        if Dif > 5:
+            print("Ultrasonic Sensor Timed out, Restarting.")
+            time.sleep(0.4)
+            Main()
+
     elapsed = stop-start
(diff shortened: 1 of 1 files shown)


dceejay:
Looks good to me. Thanks. Will adjust a bit.

PR #311, opened 2017-05-08
SNMP Set Node

mscbpi:
Added a snmp set node. My use case is a http API to SNMP for device control.

The code is a copy paste of other nodes, the main difference is in the oid format for snmp set following net-snmp lib (JSON Object):

[ { oid: ... , type:  ...,  value: .... } , { oid: ... , type:  ...,  value: .... } ] and has to be set in the message (msg.oids).

type is the integer form of net-snmp types.

### DIFF
--- a/io/snmp/snmp.html
+++ b/io/snmp/snmp.html
@@ -58,6 +58,60 @@
 </script>
 
 
+<script type="text/x-red" data-template-name="snmp set">
+    <div class="form-row">
+         <label for="node-input-host"><i class="fa fa-globe"></i> Host</label>
+         <input type="text" id="node-input-host" placeholder="localhost">
+    </div>
+    <div class="form-row">
+        <label for="node-input-community"><i class="fa fa-user"></i> Community</label>
+        <input type="text" id="node-input-community" placeholder="public">
+    </div>
+    <div class="form-row">
+        <label for="node-input-version"><i class="fa fa-bookmark"></i> Version</label>
+        <select type="text" id="node-input-version" style="width: 150px;">
+            <option value="1">v1</option>
+            <option value="2c">v2c</option>
+        </select>
+    </div>
+    <div class="form-row">
+        <label for="node-input-name"><i class="fa fa-tag"></i> Name</label>
+        <input type="text" id="node-input-name" placeholder="Name">
+    </div>
+</script>
+
+<script type="text/x-red" data-help-name="snmp set">
+    <p>Simple snmp Set node. Trigger by any input</p>
+    <p><code>msg.host</code> may contain the host.</p>
+    <p><code>msg.community</code> may contain the community.</p>
+    <p><code>msg.oids</code> must contain oids in JSON syntax according to net-snmp  [ { oid: "1.3.6.1.2.1.1.5.0", type: 4, value: "host1"} , { oid: ...} ]</p>
+    <p>The node will output <code>msg.payload</code> and <code>msg.oid</code>.</p>
+</script>
+
+<script type="text/javascript">
+    RED.nodes.registerType('snmp set',{
+        category: 'network-input',
+        color:"YellowGreen",
+        defaults: {
+            host: {value:""},
+            community: {value:"public",required:false},
+            version: {value:"1",required:false},
+            oids: {value:""},
+            name: {value:""}
+        },
+        inputs:1,
+        outputs:1,
+        icon: "snmp.png",
+        label: function() {
+            return this.name||"snmp set "+this.host;
+        },
+        labelStyle: function() {
+            return this.name?"node_label_italic":"";
+        }
+    });
+</script>
+
+
 <script type="text/x-red" data-template-name="snmp table">
     <div class="form-row">
          <label for="node-input-host"><i class="fa fa-globe"></i> Host</label>
--- a/io/snmp/snmp.js
+++ b/io/snmp/snmp.js
@@ -1,29 +1,30 @@
-module.exports = function (RED) {
+
+module.exports = function(RED) {
     "use strict";
     var snmp = require("net-snmp");
 
     function SnmpNode(n) {
-        RED.nodes.createNode(this, n);
+        RED.nodes.createNode(this,n);
         this.community = n.community;
         this.host = n.host;
         this.version = (n.version === "2c") ? snmp.Version2c : snmp.Version1;
-        this.oids = n.oids.replace(/\s/g, "");
+        this.oids = n.oids.replace(/\s/g,"");
         var node = this;
 
-        this.on("input", function (msg) {
+        this.on("input",function(msg) {
             var host = node.host || msg.host;
             var community = node.community || msg.community;
             var oids = node.oids || msg.oid;
             if (oids) {
-                node.session = snmp.createSession(host, community, { version: node.version });
-                node.session.get(oids.split(","), function (error, varbinds) {
+                node.session = snmp.createSession(host, community, {version: node.version});
+                node.session.get(oids.split(","), function(error, varbinds) {
                     if (error) {
-                        node.error(error.toString(), msg);
+                        node.error(error.toString(),msg);
                     }
                     else {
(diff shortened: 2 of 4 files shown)


knolleary:
Hi, thanks for raising this.

Please review our contribution guidelines. We prefer to discuss feature requests/ideas etc on the mailing list before they arrive as pull-requests. Otherwise you risk spending time implementing something that doesn't fit with what we may have planned, or whose design needs some tweaking before we can merge.

Nick

io/snmp/snmp.js
```diff
@@ -51,6 +51,53 @@ module.exports = function(RED) {
...
+                            }
+                            else {
+                                if (varbinds[i].type == 4) { varbinds[i].value = varbinds[i].value.toString(); }
+                                varbinds[i].tstr = snmp.ObjectType[varbinds[i].type];
+                                //node.log(varbinds[i].oid + "|" + varbinds[i].tstr + "|" + varbinds[i].value);
+                            }
+                        }
+                        msg.oid = oids;
+                        msg.payload = varbinds;
+                        node.send(msg);
```
dceejay: Why does this node need to send anything - can it not be like other output nodes and just send the data out ?

mscbpi: Just a copy paste of other snmp nodes of this package, or am I wrong ? Returns value of OIDs that were changed in this case.

bimalyn-IBM: I think @dceejay's point is that we output nodes don't generally output anything back into the flow.
Based on the [documentation](https://github.com/stephenwvickers/node-net-snmp/blob/cc4860b1b244b9822979d9822056a30e1be6cbb4/README.md), I'd modify their sample:

```javascript
var varbinds = [
    {
        oid: "1.3.6.1.2.1.1.5.0",
        type: snmp.ObjectType.OctetString,
        value: "host1"
    }, {
        oid: "1.3.6.1.2.1.1.6.0",
        type: snmp.ObjectType.OctetString,
        value: "somewhere"
    }
];

session.set (varbinds, function (error, varbinds) {
    if (error) {
        console.error (error.toString ());
    } else {
        for (var i = 0; i < varbinds.length; i++) {
            // for version 1 we can assume all OIDs were successful
            console.log (varbinds[i].oid + "|" + varbinds[i].value);
        
            // for version 2c we must check each OID for an error condition
            if (snmp.isVarbindError (varbinds[i]))
                console.error (snmp.varbindError (varbinds[i]));
            else
                console.log (varbinds[i].oid + "|" + varbinds[i].value);
        }
    }
});
```
to something like this:

```javascript
session.set (varbinds, function (error, varbinds) {
    if (error) {
        node.error (error.toString(), msg);
    } else {
        for (var i = 0; i < varbinds.length; i++) {
            // for version 1 we can assume all OIDs were successful
            // Don't do anything with this since unless there was an error, we assume success
            // console.log (varbinds[i].oid + "|" + varbinds[i].value);
        
            // for version 2c we must check each OID for an error condition
            if (snmp.isVarbindError (varbinds[i]))
                node.error (snmp.varbindError (varbinds[i]), msg);
            // Don't do anything with this since unless there was an error, we assume success
            //else
            //    console.log (varbinds[i].oid + "|" + varbinds[i].value);
        }
    }
});
```

There is no need to `node.send(msg)`, since there is no new information to be sent back into the flow.

What do you propose as the method for a user to set the varbinds? Would you want to write a helper inside of the node to convert the type `"Integer"` to `snmp.ObjectType.Integer` and `"Gauge32"` to `snmp.ObjectType.Gauge32`, etc?

mscbpi: Ok, I agree that was for error feedback but we won't manage errors within node output, so let's agree on your code.
At the moment varbinds must contain oids in JSON syntax   [ { oid: "1.3.6.1.2.1.1.5.0", type: 4, value: "host1"}  with type value in int according to net-snmp lib : 
```
var ObjectType = {
	1: "Boolean",
	2: "Integer",
	4: "OctetString",
	5: "Null",
	6: "OID",
	64: "IpAddress",
	65: "Counter",
	66: "Gauge",
	67: "TimeTicks",
	68: "Opaque",
	70: "Counter64",
	128: "NoSuchObject",
	129: "NoSuchInstance",
	130: "EndOfMibView"
}; 
```
A helper is indeed a good idea to make it more explicit.

io/snmp/snmp.html
```diff
@@ -58,6 +58,60 @@
...
+    <p>The node will output <code>msg.payload</code> and <code>msg.oid</code>.</p>
+</script>
+
+<script type="text/javascript">
+    RED.nodes.registerType('snmp set',{
+        category: 'network-input',
+        color:"YellowGreen",
+        defaults: {
+            host: {value:""},
+            community: {value:"public",required:false},
```
bimalyn-IBM: @mscbpi community is not required.

io/snmp/snmp.html
```diff
@@ -58,6 +58,60 @@
...
+            version: {value:"1",required:false},
+            oids: {value:""},
```
bimalyn-IBM: @mscbpi I'd change this to varbinds

io/snmp/snmp.js
```diff
@@ -51,6 +51,53 @@ module.exports = function(RED) {
...
+        RED.nodes.createNode(this, n);
+        this.community = n.community;
+        this.host = n.host;
+        this.version = (n.version === "2c") ? snmp.Version2c : snmp.Version1;
+        var node = this;
+
+        this.on("input", function (msg) {
+            var host = node.host || msg.host;
+            var community = node.community || msg.community;
+            var oids = msg.oids;
```
bimalyn-IBM: @mscbpi and then change this to varbinds in this file too

io/snmp/snmp.js
bimalyn-IBM: I think I would split from the node directly preceding the snmp out node.

bimalyn-IBM:
@mscbpi You'll want to pull from upstream again and take note of the use of the new `getSession` function. [Here](https://github.com/node-red/node-red-nodes/blob/master/io/snmp/snmp.js#L29) is a sample of its use.

io/snmp/snmp.js
mscbpi: OK

io/snmp/snmp.html
mscbpi: @bimalyn-IBM  you mean `oids` variable name ?

io/snmp/snmp.html
mscbpi: @bimalyn-IBM  it is in the master branch in every current node (not SnmpSNode) so I don't get the msg.community  :
```defaults: {
            host: {value:""},
            community: {value:"public",required:true},
            version: {value:"1",required:true},
            oids: {value:""},
            name: {value:""}
       }

Comment (bimalyn-IBM) [inline io/snmp/snmp.html]: Thank you for catching this. 

@dceejay I'll open a PR to fix this. It prevents the line `var community = node.community || msg.community;` from working, if we force a default value.

We should probably just set them to be `community: {value:""}`, lkie the host and oids.

File: io/snmp/snmp.html (hunk 3 above)
Comment (bimalyn-IBM) [inline io/snmp/snmp.html]: Yes, I'd change `oids` to `varbinds`. I'd recommend doing this to increase readability. If someone looks at the code signature in the node-net-snmp library it would match `session.get (oids, ...` or `session.set (varbinds, ...`. 

See [doc](https://github.com/stephenwvickers/node-net-snmp).

Comment (mscbpi) [inline io/snmp/snmp.html]: Ok, so I'll get this by the end of the week. getSession(), wrapper for object type, varbinds

File: io/snmp/snmp.html (hunk 2 above)
Comment (bimalyn-IBM) [inline io/snmp/snmp.html]: @mscbpi #324 changes the host and community to have default values, but not to require them.

```javascript
host: {value:"127.0.0.1"},
community: {value:"public"},
```
You'll need to pull from upstream again since you'll have a merge conflict.

mscbpi:
@bimalyn-IBM  follow up in #325

PR #315, opened 2017-05-10
Allos msg.mysqloptions to override the preconfigured connection information.

cgaspard:
I've created this fork to allow dynamic changing of the database mysql is connected to.   I needed this for several reasons, but primarily to handle dev ops properly, and to allow us to do things like override the connection information from within a function that can read values form an environment variable or some other place a flow can access it from.


cgaspard:
Will fix build errors and re-submit

PR #316, opened 2017-05-10
MYSQL Node: Override connection options

cgaspard:
Allows the msg.mysqloptions to override the connection pool options.

Disconnect and reconnects if any connection options were overridden.

Useful for DevOps, and using function to set connection properties from environment variables.

### DIFF
--- a/storage/mysql/68-mysql.html
+++ b/storage/mysql/68-mysql.html
@@ -65,7 +65,6 @@
     By its very nature it allows SQL injection... so <i>be careful out there...</i></p>
     <p><code>msg.topic</code> must hold the <i>query</i> for the database, and the result is returned in <code>msg.payload</code>.</p>
     <p><code>msg.payload</code> can contain an array of values to bind to the topic.</p>
-    <p><code>msg.mysqlpoolname</code> set this value to trigger mysql to change connections at runtime. It ignores the following settings otherwise.
     <p><code>msg.mysqlhost</code> can override the host using at the time message was received.
     <p><code>msg.mysqlport</code> can override the port using at the time message was received.
     <p><code>msg.mysqluser</code> can override the user using at the time message was received.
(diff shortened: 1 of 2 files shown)


dceejay:
Thanks for this.
One basic concept we try to stick to is that things specified by the user take precedence over msg properties, so msg usually only has control when the user doesn't specify something. (Or the user has an option to specify that msg has priority - but ugly)

(**Note** you should be able to specify any of the fields using environment variables using the $(MY_ENV_VAR) syntax in any field.)

cgaspard:
Hi Dave,

Is it worth me changing the logic so that it only uses the connection
options in the case that the user hasn't configured the connection then?

I was unaware of the $(VAR) syntax.   That will help me in other places
thanks.

Thanks

*Corey Gaspard*
*MyGait* *LLC*
http://www.mygait.com
*Cell: 281.794.2576*


On Wed, May 10, 2017 at 1:49 PM, Dave Conway-Jones <notifications@github.com

dceejay:
That would be the preference.

knolleary:
Just one word on this; currently none of our nodes allows their server details to be dynamically changed like this. So we need to approach this with some care as it will be establishing a precedent we must be happy with potentially applying to all similar nodes.

dceejay:
Yeah I'm not seeing an urgent need for this seeing as environment variables can be used... hmm also not sure what it would do to credentials...- don't really want those ever passed in by flow...

cgaspard:
Okay I'll hold off on the change.   The environment variables covers most of my use case.    Quick question though, how can I override credentials with environment variables?   Been searching, but haven't found any documentation on this.

dceejay:
sorry - eye off ball... the $(...) thing should work on creds also...

cgaspard:
Okay thanks.    I didn't see it working when I tried, but I may have done something wrong.   Will try again.

Issue #308, opened 2017-05-03
No support for passing options in mongdb node


thobhanifreddy:
I was trying to pass `allowDiskUse = true`. I asked question on [stackoverflow](http://stackoverflow.com/questions/43375701/pass-allowdiskusetrue-in-node-red) and realised there is no way I pass options.

thobhanifreddy:
I have edited code enough that `allowDiskUse = true` works(at least for me). Not sure if I should PR for that!

hardillb:
If you think it would be useful for others, raise a pull request and we'll have a look at it. 

Please also make sure you have read the Contributing.md  docuent in the root of this project.

thobhanifreddy:
I tried adding allowDiskUse option but my [PR](https://github.com/node-red/node-red-nodes/pull/317) fails The Travis CI build test. I am not sure what does that mean!

PR #317, opened 2017-05-17
Added allowDiskUse oprion in mongodb node

thobhanifreddy:
I have added allowDiskUse option by solving part of issue https://github.com/node-red/node-red-nodes/issues/308. 

It was also discussed at [stackoverflow](http://stackoverflow.com/questions/43375701/pass-allowdiskusetrue-in-node-red)


dceejay:
Thanks for this. Why be so specific ? Are there not other possible options to pass in ? eg batchsize ?  Why not just pass in a `msg.options` property and (in your case) set it to {allowDiskUse:true} ?
It would also make the code simpler '-)

hardillb:
Also please do not check in .DS_Store files

Issue #314, opened 2017-05-10
SNMP fails to close/destroy sessions when it errors.


Tatermen:
I have been having an issue with a SNMP heavy flow, whereby after a few hours node-red would start to crash with EMFILE errors. There are no other flows using network traffic running on this installation. 

After investigating I found that when the node-red SNMP module encounters an error it does not close the SNMP session. This means that every time an error occurs, node-red leaves a a UDP port open and listening. In my case, as I have been testing, this is leaving hundreds of open UDP ports - enough to exhaust the number of file handles. 

```
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
udp        0      0 0.0.0.0:33265           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:46577           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:59889           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:60913           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:45555           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:53747           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:39411           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:43507           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:50676           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:51700           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:47604           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:54772           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:34293           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:47605           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:36341           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:52725           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:34294           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:59894           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:54774           0.0.0.0:*                           25038/node-red
udp        0      0 0.0.0.0:35318           0.0.0.0:*                           25038/node-red
[...]

PR #318, opened 2017-05-18
Fix: SNMP fails to close/destroy sessions when it errors

bimalyn-IBM:
Fix for issue #314: SNMP fails to close/destroy sessions when it errors.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -20,5 +20,5 @@
     "requireSpacesInFunction": { "beforeOpeningCurlyBrace": true },
     //"validateParameterSeparator": ", ",
     //"validateQuoteMarks": false,
-    "maximumLineLength": 220
+    "maximumLineLength": 280
 }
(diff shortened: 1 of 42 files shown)


bimalyn-IBM:
I need to do more testing to see what happens on the socket error. What I have written so far addresses the issue that the sockets were never closing, by reusing a socket if it exists instead of spooling up more and more sockets each iteration. 

We may need to add something like this to the singleton:

```javascript
var sessions = {};

function getSession(host, community, version){
    var sessionKey = host + ":" + community + ":" + version;
    if (!(sessionKey in sessions)){
        sessions[sessionKey] = snmp.createSession(host, community, {version: version});
        /* --- START NEW SECTION  --- */
        sessions[sessionKey].on("error", function (error) {
            sessions[sessionKey].close();
            delete sessions[sessionKey];
        });
        /* --- END NEW SECTION  --- */
    }
    return sessions[sessionKey];
}
```

PR #324, opened 2017-05-31
Fix snmp community default issue

bimalyn-IBM:
Removed default value and required attribute from the community field in the snmp node.

This was preventing a user from being able to set the community value within the msg.

This issue was noticed in #311.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -20,5 +20,5 @@
     "requireSpacesInFunction": { "beforeOpeningCurlyBrace": true },
     //"validateParameterSeparator": ", ",
     //"validateQuoteMarks": false,
-    "maximumLineLength": 220
+    "maximumLineLength": 280
 }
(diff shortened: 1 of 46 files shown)


dceejay:
so now it won't work by default ? - should leave the default value... - users should remove it if they then want to override.

bimalyn-IBM:
It should still work by default. We set 

```javascript
var community = node.community || msg.community;
```

which gets thrown into

```javacript 
snmp.createSession(host, community, {version: version});
```

which creates a `Session` (see [here](https://github.com/stephenwvickers/node-net-snmp/blob/master/index.js#L548-L550) for full context):

```javascript
var Session = function (target, community, options) {
	this.target = target || "127.0.0.1";
        this.community = community || "public";
...
```

dceejay:
Agree with your last comment. I like clarity.

PR #335, opened 2017-06-13
fix deadbandEQ mode in rbe.js

jmikerq:
In the drop down menu, the deadbandEq mode is defined as "block unless value change is greater or equal to". However, in the rbe.js, the msg is only sent when the value change is equal to the gap value in deadbandEq mode.


dceejay:
Good catch - thanks

PR #336, opened 2017-06-15
Add Japanese translation file for rbe node(rbe.json)

kazuhitoyokoi:
I translated messages in rbe node into Japanese.


kazuhitoyokoi:
@dceejay Thank you for the merging! :)

PR #342, opened 2017-07-11
Adds better support for NMA

jonferreira:
Includes proper support for NMA credentials, Mustache support, all NMa API fields available, etc

Fully working after a 3 weeks test deployment on real-world instance

### DIFF
--- a/social/nma/57-nma.html
+++ b/social/nma/57-nma.html
@@ -142,39 +142,6 @@
     <p>
 
     </p>
-    <details style="padding: 0 1em;">
-    <summary>Sounds:</summary>
-        <ul style="font-size: 12px;">
-          <li>beep-crisp</li>
-          <li>beep-soft</li>
-          <li>bell-modern</li>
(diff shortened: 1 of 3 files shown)


knolleary:
Hi,
If you check out our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) you'll see we prefer to first discuss feature proposals
on the [mailing list](https://groups.google.com/forum/#!forum/node-red) or [slack team](https://nodered.org/slack). That allows us to review proposals and ensure the resulting PR and feature set are acceptable.

There are some problems with the changes you've made. First and foremost is that you've changed the node type name. That will break every single existing user of this node and is not a change we can accept.

Thanks!

jonferreira:
Sorry, first contribution I'm doing.
Indeed I've changed the node type name, my bad.
With that said I reckon it will always break compatibility given that it handles credentials differently.

dceejay:
closing as NMA have ceased operation

PR #343, opened 2017-07-12
Modify messages in e-mail node to refer to language files

kazuhitoyokoi:
To refer to language files, I changed some messages to variables and updated their language files.


kazuhitoyokoi:
Thank you @dceejay 👍

Issue #338, opened 2017-06-19
node-pisrf SRF05 bug


cyber035:
Hello I have this strange problem when using Node Red module for reading SRF04 sensor.
After some time the sensor reports some high value and gets stuck so the only way to fix it is to reboot raspberry pi or reconnect sensor...
I tried with 3 different sensors same result...

Can you add like custom value that can be set from Node Red, that specifies interval in seconds between measurements - readouts from the sensor (sending pulse and measuring echo, then wait interval in seconds for next...).

That would solve my problem completely because when I use simple py script with bigger interval between measurements, everything works..

Thanks

denisfrench:
I concur that the default interval is too short for an SRF04 on an RPi A+ (256MB) or RPi Zero W, leading to distance readings ceasing after several minutes. I don't have any other models to test with.

I have locally modified nrsrf.py to increase the interval to 2 seconds ( ie time.sleep(0.5) becomes time.sleep(2.0) ), which solved the issue for me. time.sleep(1.0) did not fix the problem.

This is an effective fix for me, but may not be suitable for time sensitive applications. Further investigation is probably warranted.

Thanks for your work on all these nodes; hundreds of thousands of us must be in your debt!

dceejay:
Added the option to be able to set the repeat time in the config

denisfrench:
Thanks for your quick fix for this @dceejay .

Looks like I spoke too soon and a longer interval just postpones the problem. That was a bit misleading on my part, but now you have a new feature :-D

I've added some debugging output locally, and this error is being thrown when the echo signal is not received:
> global name 'Main' is not defined

Give me a couple of days to play with it and I'll submit a pull request. I recommend this issue be re-opened until resolved.

PR #353, opened 2017-08-28
Improve restart upon timeout (#338)

denisfrench:
This might still need some work; I'm very rarely getting a "local variable 'stop' referenced before assignment" exception, but can't see where it is occurring.

It's happening so infrequently I can't effectively troubleshoot. It doesn't seem to be when restart() is invoked (there is none of the associated print() output to the terminal). Any assistance appreciated.

### DIFF
--- a/hardware/PiSrf/nrsrf.py
+++ b/hardware/PiSrf/nrsrf.py
@@ -19,37 +19,44 @@ SLEEP = 0.5
 
 def Measure():
     start = 0
-    realstart = 0
-    realstart = time.time()
     GPIO.output(TRIGGER, True)
     time.sleep(0.00001)
     GPIO.output(TRIGGER, False)
-    start = time.time()
-    while GPIO.input(ECHO)==0:
-        start = time.time()
-        Dif = time.time() - realstart
-        if Dif > 0.2:
-            print("Ultrasonic Sensor Timed out (pre-echo).")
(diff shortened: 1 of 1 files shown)


denisfrench:
Work-in-progress really; it seemed more reliable but I was stuck troubleshooting that occaisional error.

I've since found some inspiration here; https://electrosome.com/hc-sr04-ultrasonic-sensor-raspberry-pi/#comment-1969107914 and am working on implementing interrupts rather than polling the ECHO pin.

I'm hopeful this will significantly improve reliability. It has been my subjective impression that the whole flow was less responsive (eg missing MQTT inputs) when the nrsrf node was enabled, perhaps due to the CPU load of polling the input.

I'll get it working better and resubmit...

denisfrench:
I've converted the Measure() function to detect interrupts instead of polling the state of the ECHO pin.

This appears to have been so successful at avoiding missing the echo pulse that I have not seen the sensor timeout invoking the restart() function at all.

CPU load reduction is barely noticeable (maybe 1% on an RPi Zero W); perhaps this wasn't a problem in the first place. Otherwise this is now working fine.

dceejay:
ok - so good to merge, or any outstanding wrinkles ?

denisfrench:
Yep; seems fine at this end.

PR #362, opened 2017-10-05
Add Japanese translation file for redis node (65-redisout.json)

kazuhitoyokoi:
I translated English messages of redis node into Japanese. And I modified label tag size in html file to fit Japanese characters.


dceejay:
Thank you Yokoi-san

kazuhitoyokoi:
@dceejay You're welcome. :)

Issue #361, opened 2017-10-03
twitter node won't be able to send DMs after 1-Nov


andysc:
Twitter is removing support for the "D {username} {message}" functionality of their API.
So the advice in the info tab to send DMs won't work.
We need a new way to send a DM.

Maybe an option in the node "tweet or DM", with either a hard-wired recipient, or the ability to pass it in in a msg. variable?

fraz3alpha:
You beat me to it Andy! 

This library also uses the deprecated endpoint (https://developer.twitter.com/en/docs/tweets/post-and-engage/api-reference/post-statuses-update_with_media) and it will at some stage need to move to another one.

As some re-work will probably be required anyway, maybe the code should just parse the payload first and provide the same capability as before, but via a different implementation using the correct API calls to maintain this interface in Node-RED itself?

andysc:
oh, yes please - that would save the change from breaking a whole load of my existing apps!

hardillb:
@fraz3alpha you were already planing a pull request for your earlier Twitter node update.... ;-)

andypiper:
Actually already will not allow you to send DMs as from Oct 1. It will let you opt-in to send the string as a regular Tweet. From Nov 1, that's default, and all the options go away.

For DMs you should use the new Direct Message "beta" endpoints for [event creation](https://developer.twitter.com/en/docs/direct-messages/sending-and-receiving/api-reference/new-event).

For Tweets with media you'll need the two-step media upload, then Tweet with attached IDs, method.

eRudy:
It seems that it's already broken. I don't receive DM anymore, and get this in the console:

> HTTP Error 403: Forbidden, API message: {"errors":[{"code":151,"message":"There was an error sending your message: DM commands are no longer allowed!."}]}

:/

knolleary:
@eRudy yes - see the comment from @andypiper above.

We'll get this sorted as soon as we can... but if anyone wants to help out, just let us know!

eRudy:
Indeed @andypiper had already said that it's broken, sorry.

I would love to help, but I do not think I can write a suitable clean code :(
But I can gladly help with testing if needed !

mastix:
Hey guys,

are there any updates already?

Just a hint: I'm currently migrating from OpenHab to NodeRed and the OpenHab 2 Binding for Twitter can actually still send DMS. Maybe it's worth a look (although it's Java - but REST ist REST, right? :)).

Regards,

Sascha

[...]

PR #372, opened 2017-11-10
Bump twitter to 280 chars

### LINKED ISSUE #371: Support 280 chars tweet

hardillb:
closes #371


social/twitter/27-twitter.js
```diff
@@ -398,8 +398,8 @@ module.exports = function(RED) {
                 if (msg.hasOwnProperty("payload")) {
                     node.status({fill:"blue",shape:"dot",text:"twitter.status.tweeting"});
 
-                    if (msg.payload.length > 140) {
-                        msg.payload = msg.payload.slice(0,139);
+                    if (msg.payload.length > 280) {
```
andypiper: You'd be better off using twitter-text 2.0 (when it is released) to validate the length rather than assuming a hard-coded 280 limit. This also will not account for CJK characters which have a different weighting under the new counting algorithm https://developer.twitter.com/en/docs/developer-utilities/twitter-text

knolleary: Yeah... I see this as the quick fix ahead of a bigger rework of the node (#361)

Issue #375, opened 2017-11-18
Constrains nodegroup to numeric value


4f96a64d:
https://github.com/node-red/node-red-nodes/blob/0f83c8af54cbb22a1a5098acac470ab1e94d76ef/io/emoncms/88-emoncms.html#L34

According to this part of the emoncms code, it appears to accept alphanumeric node names. I looked around but was unable to find anything that constrained the node or nodegroup name to a numeric value.

https://github.com/emoncms/emoncms/blob/9a306ccd32140049c7ebe8d994b4a4eaf3c20e88/Modules/input/input_controller.php#L216

Use case is that some non emon* devices may want to use names that contain letters.

dceejay:
"may" want to use ?  or "do" want to use ?
Is this a real problem ?

4f96a64d:
I have been using particle photon IDs which are alphanumeric on versions that predated /^\d+$/.
PR inbound after I get rid of some rust.

dceejay:
ok - thanks.

PR #376, opened 2017-11-18
changed nodegroup field validation

4f96a64d:
Allows letters, numbers, spaces, and ( _ . - ) to more closely align with node names in emoncms.


dceejay:
Thanks for doing that

4f96a64d:
My pleasure.
I'm going to keep digging to see if there is a spec documented, but as it stands it shouldn't break anything. 
I'll close #375, thanks for your help!

PR #382, opened 2017-11-28
Multiple NIC support for WeMo devices

stephenhouser:
Update version of node-ssdp and use `explicitSocketBind` option to work with multiple network interface cards. Based on discussion in [node-ssdp issue #34](https://github.com/diversario/node-ssdp/issues/34)


hardillb:
Looks like a good idea, I just need to check on Linux as well as OSx and then I'll look at merging it

PR #383, opened 2017-11-28
node-red-node-discovery: Add status information

natcl:


natcl:
True, text is not really necessary in this instance, red just means it's not advertising.  Will also change for the ring.
I thought you could squash when merging, will have a look on my side.

knolleary:
We can squash on merge. No need for you to do it.

natcl:
Ok ! Removed status text. 
Thanks !

Issue #391, opened 2017-12-31
io/emoncms - requires update to new API syntax and options


borpin:
The emoncms input API has been changed recently and I think that the node requires an update to meet the new specification allowing full json.

borpin:
I have done a bit of work on this so that the node will now accept a valid JSON object and successfully push it to Emoncms server running 9.8.27 (which is what I needed).  I have put the files in a [gist ](https://gist.github.com/borpin/6adfc5be44e558785ae2e9fbdc8345ed) for now.

I have modified the node so that the type of data is selectable, rather than the node having to guess.  this allows for a legacy mode which should work exactly as before.

I have modified the time input to accept ISO dates or seconds.  Yes I used Date.parse which is not always ideal, to check the validity of the date.  Emoncms does more checking so it is probably safe to do.

I cannot easily check this node of other types of data.

hardillb:
Any reason not to follow the contribution guidelines and submit a pull request rather than the gist? It makes it a lot easier to see how the changes fit with the existing code.

borpin:
Part of the guidelines is to not do that until there has been a discussion :).  I'm also new to collaborating on GitHub so I'm a little concerned I may do something 'unfortunate'.  I have it in a local repository using GitHub Desktop.

dceejay:
One of the main things to check is how the changes affect existing flows... It's all too easy to all the new features to seem to be non-conflicting - but does an existing node open up with sensible defaults in any new fields etc... - but yes - happy for you to put up a PR to look at if you like.

borpin:
It is a little tricky to check how the flow will impact existing flows as I had to modify it to get it to work n the first place as the underlying server API has changed.

So this may sound daft, but how do I do a PR. Should I have created a fork first?

Paul-Reed:
Yes, create a fork of `node-red-nodes`, make your changes to the fork, test, and then submit a PR for consideration & discussion.

borpin:
Ok, I think I have managed that.  I wait and see :)

dceejay:
closing this as should now be fixed... please repoen if the new PR didn't fix (Or create a new issue)

PR #392, opened 2018-01-01
Updates to match API changes

borpin:
Update to the node to match the server API changes. An attempt has been made to leave a legacy path for existing installations that may have not updated the server side.

Allow ISO format dates.

Add output of status information.

### DIFF
--- a/io/emoncms/88-emoncms.html
+++ b/io/emoncms/88-emoncms.html
@@ -6,20 +6,20 @@
     </div>
     <div class="form-row">
         <label for="node-input-nodegroup"><i class="fa fa-object-group"></i> Node</label>
-        <input type="text" id="node-input-nodegroup" placeholder="number eg: 3">
-    </div>
-    <div class="form-row">
-        <label for="node-input-name"><i class="fa fa-tag"></i> Name</label>
-        <input type="text" id="node-input-name" placeholder="Emoncms">
+        <input type="text" id="node-input-nodegroup" placeholder="or set by msg.nodegroup">
     </div>
     <div class="form-row">
         <label for="node-input-datatype"><i class="fa fa-tag"></i> Data Type</label>
         <select id="node-input-datatype" style="width:180px;">
             <option value="legacy">Legacy Processing</option>
-            <option value="json">Simple JSON</option>
             <option value="fulljson">Valid JSON object</option>
+            <option value="json">Simple JSON</option>
             <option value="CSV">CSV</option>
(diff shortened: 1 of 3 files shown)


dceejay:
Several thoughts....
Not the end of the world but as flagged over in the Issue #391 old versions default to a blank option...
image

This can be caught by using an oneditprepare function to set it if it's undefined.
Also we usually leave the Name field as the last in the list so move the datatype up.

You have also added an output pin to the output node. We prefer nodes that "end" a flow to not have this. You can use node.status to add an indicator (and text), or node.error to throw an error - these can be caught by the status node or catch node respectively if required.

Also the info doesn't explain the difference between basic and full json options...

borpin:
Hi Thanks for the feedback. Seems I managed to get the PR part right!

1. Not sure how to implement a 'oneditprepare'; nothing immediately was thrown up by a google.
2. Output - fair enough - it helped me debug what I was doing. I understand the status, but not sure on how to implement the use of node,error.

The JSON is a bit of a peculiarity for emoncms.  Early implementations of the API had a non compliant JSON like data format that is still part of the API for legacy purposes.  The 3 types of input would be understood by someone trying to interface to the server so I think the explanation is sufficient.

Do you have some sample implementations for issues 1 & 2 that I can use?  Coding in general isn't new to me; Javascript / Node-red is.

dceejay:
Hi
re 1 - look at for example - https://github.com/node-red/node-red/blob/master/nodes/core/storage/50-file.html#L167 - 8 etc
re 2 - node.error syntax is usually node.error("some error message for the sidebar",msg)   where msg is the incoming msg that caused the error.

borpin:
Sorry, bit stuck on the oneditprepare.  Tried this code but the syntax is obviously wrong as the flow does not load.

`        oneditprepare: function() {
            if (this.datatype === undefined) {
                $("#node-input-datatype").val() = "legacy";
            }
        }
`
I'm unfamiliar with the '$' syntax and what it is actually doing. It is clear the # tells it what id to be looking at.  Trying to find a suitable reference also proved challenging.

dceejay:
it's standard jQuery - so      .val("legacy")

dceejay:
Just update this one

borpin:
Sorry, can't see an obvious way to do that :(

borpin:
Ok, think I have done that more by luck than anything else.

borpin:
Updated README to cover new functionality.

PR #393, opened 2018-01-07
Make exif node add exif data even if gps missing

jjcantillon:
The exif node was not adding exif data to the msg.exif object unless the incoming image included gps coordinates.

This did not seem consistent with the documentation for the exif node and limits the use of the node to only images that have gps data.

The node now acts as per the node description:

>Extract Exif information from JPEG images.
This node expects an incoming JPEG image buffer. If Exif data is present, it extracts the data into the msg.exif object
The node then adds location data as msg.location, should the Exif data carry this information. msg.payload remains the original, unmodified image buffer.

### DIFF
--- a/utility/exif/94-exif.js
+++ b/utility/exif/94-exif.js
@@ -73,13 +73,17 @@ module.exports = function(RED) {
                                 node.log(error.toString());
                             }
                             else {
-                                //msg.payload remains the same buffer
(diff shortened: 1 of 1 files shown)


dceejay:
Thanks for this. Good point. I'd be happy for you to not even log the fact there was no gps data.

PR #398, opened 2018-01-28
Added pass through output to email send node

colinl:
This add a message pass-through output to the email send node.  This is useful for identifying which emails have been sent and which have not, this is difficult to achieve just using the status or catch nodes.  The Telegram send node uses the same technique.


candronikos:
Hello @colinl and @coveralls.

This pull request seems very useful. Do you know when this will be available?

colinl:
@candronikos It is up the the powers-that-be whether to merge it in or not.

dceejay:
We have an item on the backlog that is a new messaging API that will allow nodes to positively acknowledge they have finished processing a message. This will be our preferred way of fixing this issue as it will fix it for all other nodes that are at the "end" of a flow. 
Visually we prefer flows to have an start and end if possible.

dceejay:
No need for this now we have the complete node in place.

PR #402, opened 2018-02-03
Android as valid platform in 88-ping.js

user-51:
[https://termux.com/](url) uses the linux ping command, but the platform string is "android". Node Red crashes when the ping node is started on Termux. This patch provides a fix by adding "android" as a valid platform.

### DIFF
--- a/io/ping/88-ping.js
+++ b/io/ping/88-ping.js
@@ -12,7 +12,7 @@ module.exports = function(RED) {
 
         node.tout = setInterval(function() {
             var ex;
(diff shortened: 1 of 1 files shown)


dceejay:
Thanks - looks good. Are you able to accept the CLA, so I can merge it ?
Otherwise I can patch by hand if you like.

user-51:
@dceejay 
> Otherwise I can patch by hand if you like.

Yes, I´d prefer that.

PR #415, opened 2018-03-16
temporary solution for direct messages

eRudy:
Added a detection of the pattern "D < user > < message >" in order to use the specific function of the twitter-ng library to send direct messages (newDirectMessage).

Sadly it's only a temporary fix, since the API will be deprecated in June 19, 2018 (https://developer.twitter.com/en/docs/direct-messages/sending-and-receiving/guides/direct-message-migration)

### DIFF
--- a/social/twitter/27-twitter.js
+++ b/social/twitter/27-twitter.js
@@ -403,8 +403,18 @@ module.exports = function(RED) {
                         node.warn(RED._("twitter.errors.truncated"));
                     }
 
-                    if (msg.media && Buffer.isBuffer(msg.media)) {
+                    if (msg.payload.slice(0,2) == "D ") {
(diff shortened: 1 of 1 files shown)


eRudy:
I corrected the code, I don't know how to solve the issue ? close this request and open a new one after a new commit ?

knolleary:
Thanks @eRudy.

If you push another commit, the PR will pick it up automatically.

However, if this is still using the deprecated API, then it doesn't buy us much. The required work is to move to the new api once and for all.

eRudy:
I understand, but should not the correction done in the twitter-ng library used by node-red ?

dceejay:
Thanks - As noted this is a short term fix... what we really need is to move to the new API

PR #426, opened 2018-03-30
Serial request

iurly:
Here goes my proposal for a serial request (initially serial inout) node.
See the [original motivation](https://groups.google.com/d/topic/node-red/38aw5MkRRwk/discussion)   and the [recent dicussion](https://groups.google.com/forum/#!topic/node-red/mxki-T6pArE) on the mailing list.
Some notes:
- It shares the same configuration with `serial in` and `serial out` for any given port. However it should not affect the behavior of the existing nodes. This means you can use as many `serial in` nodes as you want along with the new node to "sniff" on the existing serial line and `serial out` to inject mailicious/tampering data onto the serial line skpping the queueing mechanism.
- I'm not sure the approach for duplicating a message is correct (Object.assign()), perhaps I'm potentially creating duplicate _msgid?
- I'm keeping a reference to the node (I mean Node-RED's node) within the internal queue to track which node the request originally came from, and hack the emitter paradigm to perform some filtering, I'm not sure this is 100% safe.
- ~most of the code is a copy from both serial in serial out, so I'm well aware it could use some refactoring -- feedback welcome!~ _recently refactored_
- every incoming message should eventually produce an outgoing message, I came up with the "status" field using some strings, perhaps there's a better solution
- this implementation shows some status on the node, which might turn annoying to some people. If accepted, it could use some improvement in that it does not distinguish between "message enqueued (X messages waiting)" and "message sent"
- ~ideally the timeout should also be configurable through the configuration node~ _DONE_.
- ~some other potential features (like delayed transmission) are stricken out from the documentation as they're not implemented~ _REMOVED from the documentation_.

**New feature**: as part of this PR there's also a new msg split logic: "silent line timeout". In this case the timeout is re-started at each received character.
Any feedback welcome! Thanks in advance!

### DIFF
--- a/io/serialport/25-serial.html
+++ b/io/serialport/25-serial.html
@@ -82,6 +82,62 @@
     });
 </script>
 
+<script type="text/x-red" data-template-name="serial request">
+    <div class="form-row node-input-serial">
+        <label for="node-input-serial"><i class="fa fa-random"></i> <span data-i18n="serial.label.serialport"></span></label>
+        <input type="text" id="node-input-serial">
+    </div>
+    <div class="form-row">
+        <label for="node-inputoutput-name"><i class="fa fa-tag"></i> <span data-i18n="node-red:common.label.name"></span></label>
+        <input type="text" id="node-input-name" data-i18n="[placeholder]node-red:common.label.name">
+    </div>
+</script>
+
+<script type="text/x-red" data-help-name="serial request">
+    <p>Provides a connection to a request/response serial port.</p>
+    <p>This node behaves as a tightly coupled combination of <code>serial in</code> and <code>serial out</code> nodes,
+    with which it shares the configuration.</p>
+    <p>Send the request message in <code>msg.payload</code> as you would do with a <code>serial out</code> node.
+    The message will be forwarded to the serial port following a strict FIFO (First In, First Out) queue, waiting for a single response before transmitting the next request.
+    Once a response is received (with the same logic of a <code>serial in</code> node), or after a timeout occurs, a message is produced to the output (see Outputs below),
+    with <code>msg.payload</code> containing the received response (or missing in case if timeout) and all other fields preserved.</p>
+    <h3>Inputs</h3>
+    <ul>
+    <li>
+    <code>msg.timeout</code> is the timeout (in ms) after which the incoming message is propagated to the output with <code>msg.status</code> set to <code>"ERR_TIMEOUT"</code> and missing payload.
+    If not present, the default value is 10000 (10s).
+    </li>
+    <li>Any other field will be propagated to the output so to allow to match responses with requests.</li>
+    </ul>
+    <h3>Outputs</h3>
+    <ul>
+    <li><code>msg.payload</code> is the response. If no response occured, this field is removed.</li>
+    <li><code>msg.status</code> is <code>"OK"</code> in case a response is received, or <code>"ERR_TIMEOUT"</code> if a timeout occurs.</li>
+    <li>Any other field coming from the input will be preserved.</li>
+    </ul>
+</script>
+
+<script type="text/javascript">
+    RED.nodes.registerType('serial request',{
+        category: 'function',
+        defaults: {
+            name: {name:""},
+            serial: {type:"serial-port",required:true}
+        },
+        color:"BurlyWood",
+        inputs:1,
+        outputs:1,
+        icon: "serial.png",
+        align: "left",
+        label: function() {
+            var serialNode = RED.nodes.node(this.serial);
+            return this.name||(serialNode?serialNode.label().split(":")[0]:this._("serial.label.serial"));
+        },
+        labelStyle: function() {
+            return this.name?"node_label_italic":"";
+        }
+    });
+</script>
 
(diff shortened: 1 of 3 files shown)


io/serialport/25-serial.html
```diff
@@ -82,6 +82,66 @@
...
+    with which it shares the configuration.</p>
+    <p>Send the request message in <code>msg.payload</code> as you would do with a <code>serial out</code> node.
+    The message will be forwarded to the serial port following a strict FIFO (First In, First Out) queue, waiting for a single response before transmitting the next request.
+    Once a response is received (with the same logic of a <code>serial in</code> node), or after a timeout occurs, a message is produced to the output (see Outputs below),
+    with <code>msg.payload</code> containing the received response (or missing in case if timeout) and all other fields preserved.</p>
+    <h3>Inputs</h3>
+    <ul>
+    <li>
+    <code>msg.timeout</code> is the timeout (in ms) after which the incoming message is propagated to the output with <code>msg.status</code> set to <code>"ERR_TIMEOUT"</code> and missing payload.
+    If not present, the default value is 10.000 (10s). <del>If set to zero, the message is immediately propagated to the output with <code>msg.status</code> set to <code>"OK"</code> and missing payload</del>.
```
dceejay: these <del> tags render as strikethrough....  not needed.

io/serialport/25-serial.html
```diff
@@ -82,6 +82,66 @@
...
+    </li>
+    <li><del><code>msg.delay</code> is the delay to wait before the message is sent to the serial device.
```
dceejay: same again....

io/serialport/25-serial.html
```diff
@@ -82,6 +82,66 @@
...
+    Default value is 0.</del></li>
+    <li><del>Set <code>msg.reset</code> to `true` to clean up the existing queue (in this case <code>msg.payload</code>
```
dceejay: and again

dceejay:
Hi @iurly - any chance to fix this up so we can merge it ?

iurly:
Busy week, sorry! I'll pick this up again early next week. While at it, I was wondering if the unbounded queue is really a good idea and if we should do something about it... Thoughts?

dceejay:
Indeed (re queue) - other nodes do allow you to specify depth in settings.js if you want (but don't require that you do). eg split, sort etc - maybe something like `serialRequestBufferMaxLength`

iurly:
Any idea why my build is suddenly failing?

dceejay:
So ... finally get a chance to look - seem good so far.
couple of nits to clean up..
image
should be made wider so the text fits.

Also the input node sets msg.port so you know where it comes from... is this the case here ? (though I guess as it's a request you know where you sent it so shouldn't need to know.)

The node.status messages should be translatable - using the locales/en-US/messages.json file rather than being hardcoded.

But yes - so far seems good.

io/serialport/25-serial.js
```diff
@@ -358,7 +358,13 @@ module.exports = function(RED) {
                                 // do the timer thing
                                 if (spliton === "time") {
                                     // start the timeout at the first character
-                                    if (!obj.tout) {
+                                    if ( i == 1 ) {
```
dceejay: This mode should only be enabled when in split on time from last character mode - not the current timeout mode.

iurly: Hmm... keep in mind this version does not include the "new" behavior.
It behaves exactly as before: it starts the timeout the at the first character. It just uses a different test to establish it's the first character, as obj.tout could have been set for a different purpose already.
Anyway, I'm about to push the commit for the new behavior.

dceejay:
Looks good
any further thoughts/changes/tidying before we final review / merge ?

dceejay:
HI - any further thoughts/changes/tidying before we final review / merge ?

dceejay:
I think this is the most "official" list - https://nodejs.org/api/errors.html#nodejs-error-codes
but I don't think they are sacred if there isn't one close enough.

iurly:
I see, pretty interesting reading.
But I was talking about some standard way to notify of errors through `msg`s within Node-RED, if any.
`http request` reports some `msg.statusCode`, for instance (numeric, I guess).
I guess we may stick with the existing implementation then, it shouldn't sound totally insane.

dceejay:
ah right. yes. the "normal" Node_RED way is to pass errors through node.error call instead - but again that's not what you want. I going numeric you could follow http... then again bash just goes 0 for  and anything else for a return code depending on severity... so ...
So yes leave it as-is I think.

There is an outstanding work-item on the backlist for how to signal success more generally (for nodes that have no output).

dceejay:
OK - all merged. I'm going to leave it here for a week or so so we can more thoroughly test it before pushing to npm.

iurly:
Cool! :-)
Do you have any instructions on how to install it for anyone who'd like to try it out?

dceejay:
OK - I have created a beta tag so 

    cd ~/.node-red
    npm i node-red-node-serialport@beta

should do it

yfhaotian:
Hello, do you know how serial request uses msg.baudrate to configure the baud rate?

Issue #429, opened 2018-04-09
io/emoncms - User reports Timeout period is too short


borpin:
On the Emoncms Forum a user reported that the newly introduced HTTP Timeout appeared to be too short.

https://community.openenergymonitor.org/t/timeout-object-object-error-in-node-red/6990

This timeout was only intended to prevent gross errors and allow the capture of an event whereby no response was received.  It was suggested a time of 60S would be more appropriate.  It has been noted in other circumstances that there are users who communicate over high latency links so this is not surprising.

https://github.com/borpin/node-red-nodes/blob/d28c5291c0a1b22eb381498e0c87e39a173a3c34/io/emoncms/88-emoncms.js#L141

User reported that changing the value solved the error.

In addition, the error message needs to be amended so the message is better than “timeout: [object Object]”.

Suggested solution is to increase this value to 6000 and modify error message.  Happy to create PR.

dceejay:
Yes please - PR always good if you have one.

PR #430, opened 2018-04-10
io/emoncms - Modify HTTP Timeout

borpin:
Issue #429

Modify timeout value from 1000 to 6000

Modify error message.


borpin:
@dceejay - Done.

dceejay:
Thanks

PR #440, opened 2018-04-30
upgrade nodemailer; closes #439

### LINKED ISSUE #439: upgrade to non-deprecated version of nodemailer
```
nodemailer@1.11.0: All versions below 4.0.1 of Nodemailer are deprecated. See https://nodemailer.com/status/
```

boneskull:
- update documentation(s)
- add/update `package-lock.json`(s)
- add `engines` field(s)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ addons:
 node_js:
   - "8"
   - "6"
-  - "4"
 before_script:
   # Remove the './node_modules/.bin:' entry, see https://github.com/travis-ci/travis-ci/issues/8813
   - export PATH=`echo ${PATH} | sed -re 's,(^|:)(./)?node_modules/.bin($|:),\1,'`
--- a/social/email/README.md
+++ b/social/email/README.md
@@ -12,6 +12,8 @@ You will need valid email credentials for your email server.
 Install
 -------
 
+**Node.js v6 or newer is required.**
+
 Run the following command in your Node-RED user directory - typically `~/.node-red`
 
         npm i node-red-node-email
--- a/social/email/package.json
+++ b/social/email/package.json
@@ -1,6 +1,6 @@
 {
   "name": "node-red-node-email",
-  "version": "0.1.29",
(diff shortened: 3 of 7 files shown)


boneskull:
the failing build is my way of doing TDD

social/email/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.1.29",
```
dceejay: As per discussions we're going to need to major bump this.

.travis.yml
```diff
@@ -12,7 +12,6 @@ addons:
 node_js:
   - "8"
   - "6"
```
dceejay: Does it actually fail on 4 - or just not supported on 4 ? (ie have they actively broken it ?)

social/email/README.md
```diff
@@ -12,6 +12,8 @@ You will need valid email credentials for your email server.
 Install
 -------
 
+**Node.js v6 or newer is required.**
+
```
dceejay: Think we are going to need more explicit instructions how to upgrade - as it's a core node it's not installed  in ~/.node-red so need sudo etc... - or at least sudo to remove old then install this etc... - but I guess we can finness that before release.

social/email/package.json
boneskull: I don't usually bump versions in PRs; I let the maintainer decide.  But if you want me to bump it, I can.

.travis.yml
boneskull: It's actually broken

social/email/package.json
dceejay: true - but a) as it's you and b) in this case I want to make sure we get no possible crossover with the 0.1 code.

social/email/README.md
boneskull: Any such upgrade guide is going to depend on how Node.js 4 was installed originally, and what platform the user is on.  If using `apt`, one would hope that upgrading would automatically remove the old version...

PR #466, opened 2018-07-19
Add support for batch of SQL statements

tmdoit:

## Checklist

### DIFF
--- a/storage/sqlite/sqlite.js
+++ b/storage/sqlite/sqlite.js
@@ -61,6 +61,23 @@ module.exports = function(RED) {
                         }
                     }
                 }
+                if (this.sqlquery == "batch") {
+                    if (typeof msg.topic === 'string') {
+                        node.mydbConfig.db.exec(msg.topic, function(err) {
+                            if (err) { node.error(err,msg);}
+                            else {
+                                delete msg.payload;
+                                node.send(msg);
+                            }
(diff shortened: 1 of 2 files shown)


storage/sqlite/sqlite.js
```diff
@@ -61,6 +61,23 @@ module.exports = function(RED) {
...
+                        node.mydbConfig.db.exec(msg.topic, function(err) {
+                            if (err) { node.error(err,msg);}
+                            else {
+                                delete msg.payload;
+                                node.send(msg);
+                            }
+                        });
+                    }
+                    else {
+                        if (typeof msg.topic !== 'string') {
```
dceejay: Is this necessary (as it's already in the else from above ?)

storage/sqlite/sqlite.js
```diff
@@ -61,6 +61,23 @@ module.exports = function(RED) {
                         }
                     }
                 }
+                if (this.sqlquery == "batch") {
+                    if (typeof msg.topic === 'string') {
+                        node.mydbConfig.db.exec(msg.topic, function(err) {
+                            if (err) { node.error(err,msg);}
+                            else {
+                                delete msg.payload;
```
dceejay: I think it should return an empty array on success (like other modes). So at least it contains something...

Issue #467, opened 2018-07-23
support HTML-mode in Pushover node


cowchimp:
Hi.

Would you be interested in a Pull-Request that will allow setting `html` mode on an outgoing Pushover notification?
It's a mode they now support which means that if you include HTML in your notification payload it will be rendered accordingly.
You can read about it in their [docs](https://pushover.net/api#html).

If you are interested, would you like me to include this is a radio-button in the node UI, or just as an option which you can enable by setting the `html` property of the `msg` passed to the Pushover node?

Thanks.

dceejay:
Sounds good. Checkbox or select box preferred. Default to existing so existing flows continue to work please.

knolleary:
Sounds reasonable - probably should be an option to enable html handling.

For future reference, *please* read the issue template you are shown when raising issues. You would have seen we prefer feature requests to be made via the forum, rather than the issue tracker.

PR #468, opened 2018-07-23
support HTML-mode in Pushover node #467

cowchimp:

Adds an option to toggle HTML mode as discussed in #467.
Tested it with open on\off.
Is backward compatible (HTML mode is off)

## Checklist


cowchimp:
Hi.
Build now passes after merging upstream into my fork.
I made the changes as we discussed in the issue.
Would you like to change anything else or is this okay to merge?
Thanks.

social/pushover/57-pushover.js
```diff
@@ -35,6 +36,7 @@ module.exports = function(RED) {
             var sound = this.sound || msg.sound || null;
             var url = this.url || msg.url || null;
             var url_title = this.url_title || msg.url_title || null;
+            var html = this.html || msg.html || false;
```
dceejay: If the user sets the flag to false then it should be honoured. This won't do that - as the this.html will evaluate false and then skip to the next option (msg.html)... so msg.html could always override it.

cowchimp: Good point. I removed the option to toggle HTML mode from the incoming msg.
Now that I've given it some more thought, I can't think of a reasonable case where you would need to make this dynamic and change based on the `msg` anyway.

dceejay:
Thanks for the PR

cowchimp:
@dceejay `node-red-nodes` depends on `pushover-notifications: ~0.2.3`.
Since that version `pushover-notifications` has released only two version: `0.2.4` and `1.0.0`.
I tested with `0.2.4`.
As for what changed in `1.0.0`, I didn't find a proper changelog at
https://github.com/qbit/node-pushover
but judging from the commit it looks like it was just a caution step because they applied a mass automated reformating to the codebase (using the `standardjs` formatter), and not an API change by choice.
https://github.com/qbit/node-pushover/commit/eeb6d2a118c9c5513d20096fc4232acb5adcf662?w=0#diff-04c6e90faac2675aa89e2176d2eec7d8

dceejay:
... so....  did you test with 1.0 ? - what is your recommendation ?
(Not used pushover in ages... and don't have time to get deep into it now.. so trusting to you :-)

cowchimp:
I just tested with `1.0.0`.
Technically it does work and it's backward compatible, but in my opinion I see no reason to update.

dceejay:
ok - will go with 0.2.4

cowchimp:
Cheers. Thanks.

PR #476, opened 2018-08-02
Update to the latest mysql package

tmdoit:
Updating from mysql 2.13 to 2.16 fixed for me a problem with error msg "Error: pool is closed" produced every few seconds in NR debug window.

## Checklist


dceejay:
Thanks. Will update when back from vacation. For new installs the ~ should install that version anyway

PR #480, opened 2018-08-22
Node red mail usetls

Schreck1985:

## Proposed changes
Have the option to disable TLS checking for sending mails. See (https://discourse.nodered.org/t/node-red-node-email-tls-check/1853)

## Checklist

### DIFF
--- a/social/email/61-email.js
+++ b/social/email/61-email.js
@@ -30,7 +30,7 @@ module.exports = function(RED) {
         this.outserver = n.server;
         this.outport = n.port;
         this.secure = n.secure;
-        this.tls= n.tls;
+        this.tls = true;
         var flag = false;
         if (this.credentials && this.credentials.hasOwnProperty("userid")) {
             this.userid = this.credentials.userid;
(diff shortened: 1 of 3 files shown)


social/email/61-email.js
```diff
@@ -55,7 +56,8 @@ module.exports = function(RED) {
         var smtpOptions = {
             host: node.outserver,
             port: node.outport,
-            secure: node.secure
+            secure: node.secure,
+            tls: {rejectUnauthorized: node.tls}
```
dceejay: just checking - letting this be null or undefined is OK ?

Schreck1985: Maybe we should check for null or at least set default value of true. Sorry, it is my very first touch to JavaScript.

dceejay: Yes - (hey no problem, glad you are learning). Yes - further up set it true by default - then test if n.tls === false - so only set it false if explicitly set to false by the user.

Schreck1985:
Damn, I don't know what is wrong with mit GIT config. Sorry for that.

dceejay:
Thanks for persevering

PR #504, opened 2018-11-14
Use connection pool for query

svrist:
This allows the nodes to use multiple connections to the same database
which avoid serializing access.

## Proposed changes
We have a node-red flow with multiple API endpoints GET /api/this, GET /api/that etc. that all access the database. Most often simple "SELECT something FROM table"

We observed that if one of those endpoints did a query that took a bit longer (seconds) - all other endpoints accessing the MySQL would hang until that slow request finished.

Looking at the mysql `SHOW PROCESSLIST` it shows only a single connection from node-red
``` 
+-----+-------------+-----------------+------+---------+------+--------------------------+------------------+----------+
| Id  | User        | Host            | db   | Command | Time | State                    | Info             | Progress |
+-----+-------------+-----------------+------+---------+------+--------------------------+------------------+----------+
|   9 | nodered     | localhost:37780 | te   | Sleep   |   18 |                          | NULL             |    0.000 |
+-----+-------------+-----------------+------+---------+------+--------------------------+------------------+----------+

``` 

And when the slow query is running that connection is then serving that query.

Looking at the node-red-node-mysql I looks like a mysql connection pool with allowed 25 connections is setup - but then only one connection is collected and used in perpetuity

I tried patching the node_modules/ installed version of node-red-node-mysql/68-mysql.js  to use the `pool` to do `query` instead of `connection` as the mysqljs/mysql documentation mentions `pool.query` as a short hand for `getConnection, query, release`.

Running  with that I see multiple connections, but not neccesarily all 25. I figure it's smart about only allocation new when needed. 
```
+-----+-------------+--------------------+------+---------+------+--------------------------+------------------+----------+
| Id  | User        | Host               | db   | Command | Time | State                    | Info             | Progress |
+-----+-------------+--------------------+------+---------+------+--------------------------+------------------+----------+
|  94 | nodered     | xxx.xx.xx.xx:48062 | beta | Sleep   |  199 |                          | NULL             |    0.000 |
|  95 | nodered     | xxx.xx.xx.xx:48064 | beta | Sleep   |   48 |                          | NULL             |    0.000 |
|  99 | nodered     | xxx.xx.xx.xx:48164 | beta | Sleep   |   18 |                          | NULL             |    0.000 |
| 100 | nodered     | xxx.xx.xx.xx:48200 | beta | Sleep   |   78 |                          | NULL             |    0.000 |
+-----+-------------+--------------------+------+---------+------+--------------------------+------------------+----------+
``` 
And it alleviates the hanging API endpoints when running concurrently with slow queries.

*Caveats exist* as this will subtly change the behaviour under some circumstances, specifically if you are using flows that have multiple concurrent accesses to the database. Before this change all database access would been serialized (in some order) and executed one-after-the-other. This change will allow them to run concurrently with database concurrency issues as in any other database-using-application. For example aborted queries due to deadlocks etc.

## Checklist

https://discourse.nodered.org/t/mysqldbnode-reuses-the-same-connection-for-all-access/4856
- [?] I have added suitable unit tests to cover the new/changed functionality

I've added just as many as there already existed for the storage modules :)

### DIFF
--- a/storage/mysql/68-mysql.js
+++ b/storage/mysql/68-mysql.js
@@ -121,7 +121,7 @@ module.exports = function(RED) {
                     if (typeof msg.topic === 'string') {
                         //console.log("query:",msg.topic);
                         var bind = Array.isArray(msg.payload) ? msg.payload : [];
-                        node.mydbConfig.connection.query(msg.topic, bind, function(err, rows) {
+                        node.mydbConfig.pool.query(msg.topic, bind, function(err, rows) {
                             if (err) {
                                 node.error(err,msg);
                                 node.status({fill:"red",shape:"ring",text:"Error"});


svrist:
I just tried it with one of our complicated flows and it gave a set of deadlock issues needing to straighten out. So I'm not sure this is a overall drop-in change

dceejay:
Thanks for letting us know. Definitely would like to have something like this to help improve performance... but yes... not at the cost of breaking things in a more complicated manner :-)
Will hold off merging while you tinker.

OverWorldD:
I'm getting "pool is closed" errors sometimes. Is that because of serialization timing issues, would they be fixed with this change?

OverWorldD:
Hi! Any updates on this, I'm getting swamped with pool is closed errors here... I was thinking this might be a solution

dceejay:
Hi @svrist - any more thoughts ?

sbv-trueenergy:
I can't generally recommend this solution. It's not always backwards compatible. A minor test to see if you would benefit is to add the same database multiple times and at least not reuse the same connection across all nodes doing db actions. It's a very simple patch so npm also supports local patches if you are keen to try it out.

I guess it would need to be a configurable opt in thing if it should go into the main repo but I'm out of time to propose PR for it

PR #542, opened 2019-05-21
add to email threading headers

domenzain:
## Types of changes

## Proposed changes
Discussed in: https://discourse.nodered.org/t/use-reply-headers-for-email-threading/11364

> Currently it is possible to have e-mails as inputs or outputs of flows.
> 
> Some headers can be set directly, like `To` and `CC`, but not all: specifically, `In-Reply-To` and `References` should be updated for email threading to work as expected.
> 
> This would allow for flows where an email by Node-RED can be clearly marked as a reply to the email that initially triggered the flow.
> 
> An easy way of implementing this is to add those fields to the existing `nodemailer` interface, as it implements them already.

Add email threading headers that nodemailer already supports:
- msg.replyTo
- msg.inReplyTo
- msg.references

and update documentation in available locales

## Checklist

- [x] For non-bugfix PRs, I have discussed this change on the forum.

### DIFF
--- a/social/email/61-email.html
+++ b/social/email/61-email.html
@@ -65,7 +65,7 @@
     <p>Sends the <code>msg.payload</code> as an email, with a subject of <code>msg.topic</code>.</p>
     <p>The default message recipient can be configured in the node, if it is left
     blank it should be set using the <code>msg.to</code> property of the incoming message. If left blank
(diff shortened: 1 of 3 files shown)


dceejay:
Looks good - Thanks

domenzain:
It is a pleasure to contribute to such reactive communities.
Thank you.

PR #546, opened 2019-05-28
use latest mailparser API

domenzain:
## Types of changes

## Proposed changes

The API diverged between the original mailparser and mailparser-mit, which causes the email node to silently fail to emit any of the fetched emails.

This pull request uses the latest mailparser API.

## Checklist


domenzain:
@dceejay,

This slipped through on the last bugfix because there is no evident error other than e-mails not being parsed and sent as messages...

dceejay:
Thanks

PR #547, opened 2019-05-29
email optionally uses criteria from message

domenzain:
## Types of changes

## Proposed changes

Optionally use `msg.criteria` to set the IMAP criteria.
This allows for generation of detailed queries through flows when triggering an email
fetch in IMAP.

Detailed criteria is available at the underlying library's README:
https://github.com/mscdex/node-imap

A simple example in a function node to fetch a specific message again, even though it might already be "Seen":
```javascript
criteria = [['HEADER', 'message-id', msg.header['message-id']]]
msg = {}
msg['criteria'] = criteria
```
This coupled with a fetch and a different disposition allows a selective deleting of emails after some processing.

Discussion:
https://discourse.nodered.org/t/provide-search-criteria-to-imap-on-email-node-on-trigger/11642

## Checklist

- [x] For non-bugfix PRs, I have discussed this change on the forum.

### DIFF
--- a/social/email/61-email.js
+++ b/social/email/61-email.js
@@ -353,7 +353,7 @@ module.exports = function(RED) {
                     }
                     else {
                         var criteria = ((node.criteria === '_msg_')?
-                                        (msg.criteria || []):
+                                        (msg.criteria || ["UNSEEN"]):
                                         ([node.criteria]));
                         imap.search(criteria, function(err, results) {
                             if (err) {
--- a/social/email/61-email.js
+++ b/social/email/61-email.js
@@ -352,7 +352,10 @@ module.exports = function(RED) {
                         return;
                     }
                     else {
-                        imap.search([ node.criteria ], function(err, results) {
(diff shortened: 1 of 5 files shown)


social/email/61-email.js
```diff
@@ -352,7 +352,10 @@ module.exports = function(RED) {
                         return;
                     }
                     else {
-                        imap.search([ node.criteria ], function(err, results) {
+                        var criteria = ((node.criteria === '_msg_')?
+                                        (msg.criteria || []):
```
dceejay: should the default be ALL ?

domenzain: It could be the same default as the new node.
And it is still possible to pass `[]` anyhow.

```suggestion
                                        (msg.criteria || ["UNSEEN"]):
```

dceejay: yes probably better if it does the default if not set.

domenzain: Applied.
I did test locally.

dceejay:
Thanks. again.

PR #552, opened 2019-07-02
Visually separated the list of control values from the example.

hepcat72:
This will aid people who are quickly skimming the readme. There wasn't anything wrong in the doc. Everything would be clear if the reader is reading thoughtfully. But if they're skimming and looking for what to supply to the node for the first time, their eye is drawn to the JSON example.  Since many nodes accept JSON input, and supplying `{"state": "off"}`, `{"state": 0}`, or `{"state": 1}` works, it might take them awhile to realize their oversight.

- [x] Bugfix (non-breaking *documentation* change which fixes an issue)

## Checklist

- [x] For non-bugfix PRs, I have discussed this change in another issue.

### DIFF
--- a/hardware/wemo/README.md
+++ b/hardware/wemo/README.md
@@ -21,10 +21,14 @@ from opening the config dialog and selecting the node you want.
 
 The node accecpts the following inputs
 
- * Strings on/off
- * integers 1/0
- * boolean true/false
- * an Object like this (lights only & color control is still work in the progress) 
+ * A single value, such as
+     * Strings `on`/`off`
+     * integers `1`/`0`
+     * boolean `true`/`false`
+```
+on
+```
+ * A JSON Object like this (lights only & color control is still work in the progress) 
  ```
     {
       state: 1,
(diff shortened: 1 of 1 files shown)


hardware/wemo/README.md
```diff
@@ -21,10 +21,14 @@ from opening the config dialog and selecting the node you want.
...
- * Strings on/off
- * integers 1/0
- * boolean true/false
- * an Object like this (lights only & color control is still work in the progress) 
+ * A single value, such as
+     * Strings `on`/`off`
+     * integers `1`/`0`
+     * boolean `true`/`false`
+```
+on
```
dceejay: on ? or did you mean or ?

hepcat72: It's an example of what to send to the node, so `on`. I admit it's weird that there are sub-items between the item preceding the example. Let me see if I can improve it...

dceejay: ok - so it should be titled "object"  (like integer boolean etc) - then have the example.

dceejay: actually - or without the blockquotes would work fine.

Also while you are at it... - just before the list maybe make the line
The node accecpts the following inputs   - into - The node accepts the following `msg.payload` inputs

PR #557, opened 2019-07-19
Tail node add input filename

zinen:

## Proposed changes

Added filename parameter as an input, to allow for dynamically change the files being monitored. 
I needed it, and now others can use it too.
Is backwards compatible with existing flows.

## Checklist

### DIFF
--- a/storage/tail/28-tail.html
+++ b/storage/tail/28-tail.html
@@ -47,7 +47,8 @@
             name: {value:""},
             filetype: {value:"text"},
             split: {value:"[\\r]{0,1}\\n"},
-            filename: {value:""}
+            filename: {value:""},
+            inputs: {value:1}
         },
         color:"BurlyWood",
         inputs:1,
@@ -64,6 +65,14 @@
                 if (this.value === "text") { $("#node-tail-split").show(); }
                 else { $("#node-tail-split").hide(); }
             });
-        }
+        },
+        oneditsave: function() {
+            var that = this;
+            if($("#node-input-filename").val()===""){
+                that.inputs = 1;
+            } else {
+               that.inputs = 0;
+           }
+       }
     });
 </script>
--- a/storage/tail/28-tail.html
+++ b/storage/tail/28-tail.html
@@ -31,7 +31,7 @@
     <dl class="message-properties">
         <dt class="optional">filename <span class="property-type">string</span></dt>
         <dd>If not configured in the node, this optional property sets the name of the file to be tailed.
-        If an empty string <code>""</code> is reviced, the tailing will stop.</dd>
+        If an empty string <code>""</code> is received, the tailing will stop.</dd>
     </dl>
     <h3>Outputs</h3>
     <ul>
--- a/storage/tail/28-tail.html
+++ b/storage/tail/28-tail.html
@@ -27,6 +27,12 @@
 <script type="text/x-red" data-help-name="tail">
     <p>Tails (watches for things to be added) to the configured file.</p>
     <p>Note: On Windows you may need to close the file between writes for any updates to be registered.</p>
+    <h3>Input</h3>
+    <dl class="message-properties">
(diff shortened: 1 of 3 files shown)


storage/tail/28-tail.html
```diff
@@ -27,6 +27,12 @@
 <script type="text/x-red" data-help-name="tail">
     <p>Tails (watches for things to be added) to the configured file.</p>
     <p>Note: On Windows you may need to close the file between writes for any updates to be registered.</p>
+    <h3>Input</h3>
+    <dl class="message-properties">
+        <dt class="optional">filename <span class="property-type">string</span></dt>
+        <dd>If not configured in the node, this optional property sets the name of the file to be tailed.
+        If an empty string <code>""</code> is reviced, the tailing will stop.</dd>
```
dceejay: Received

storage/tail/28-tail.html
```diff
@@ -41,10 +47,10 @@ <h3>Outputs</h3>
             name: {value:""},
             filetype: {value:"text"},
             split: {value:"[\\r]{0,1}\\n"},
-            filename: {value:"",required:true}
+            filename: {value:""}
         },
         color:"BurlyWood",
-        inputs:0,
+        inputs:1,
```
dceejay: I think it may be possible to make this dynamic based on filename field being empty or not

storage/tail/28-tail.js
```diff
@@ -46,7 +47,29 @@ module.exports = function(RED) {
...
+        if (node.filename !== "") {
+            node.status({});
+            fileTail();
+        } else {
+            node.status({ fill: "grey", text: "paused" });
+            node.on('input', function (msg) {
+                if (!msg.hasOwnProperty("filename")) {
+                    node.error("Missing filename on input");
+                }else if (msg.filename === node.filename) {
+                    node.warn("No change of filename input");
```
dceejay: Not sure this needs to be a warm. If you have paused it for some reason you would want this to restart

storage/tail/28-tail.js
```diff
@@ -37,6 +37,7 @@ module.exports = function(RED) {
                 });
 
                 node.tail.on("error", function(err) {
+                    node.status({ fill: "red",shape:"ring", text: "error" });
```
dceejay: I think we may need to make all strings i18n translated

zinen: Fixed with 4256a76 and ac40455

storage/tail/28-tail.html
zinen: Fixed with d403746

storage/tail/28-tail.html
zinen: I tried a bunch of stuff to get this working. Didn't work in the end.
Even tried implementing parts from here:
https://github.com/node-red/node-red-nodes/blob/master/social/email/61-email.html#L326

storage/tail/28-tail.js
zinen: I'made it as a warning because the user should not think of it as needing an injection to do the tailing.
It works so feeding the node with an empty string will stop the tailing, and feeding it again with at new value/file would start tailing the new file. Feeding the node with the same value twice serves therefor no purpose, since the tail is already active.

storage/tail/28-tail.html
dceejay: Yes the email node should be an example that works.

zinen: Fixed with 82edc64.
Input are now dynamic changed according to filename defined or not.

scelle:
hi, 

I'm a bit confused about this added input.filename property.   I wanted to use this option, but there's no input on the tail node.  So how do I set this property then? Am I missing some basic node-red functionality I wasn't aware of?

dceejay:
If you don’t set it in the node then save the config the input should appear.

Issue #567, opened 2019-08-12
node-red-node-pi-mcp3008 does not close properly


edreanernst:
### Which node are you reporting an issue on?
node-red-node-pi-mcp3008

### What are the steps to reproduce?
1. Drop this flow:

`[{"id":"ffcacc48.388e","type":"inject","z":"f362c22.31c484","name":"","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":100,"y":1280,"wires":[["fd2f15c3.ad28b8"]]},{"id":"fd2f15c3.ad28b8","type":"pimcp3008","z":"f362c22.31c484","name":"","dev":"3008","pin":0,"dnum":0,"bus":0,"x":290,"y":1280,"wires":[["3b074c74.8a2214"]]},{"id":"3b074c74.8a2214","type":"debug","z":"f362c22.31c484","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","x":490,"y":1280,"wires":[]}]
`

2. Deploy
3. Restart flows
4. Click on inject node.

### What happens?
An error is raised : "Error: EPERM, device closed, operation not permitted"

If the nodered service is restarted it works again until the flows are restarted.

### What do you expect to happen?
The node should not generate errors when it is triggered after a restart of the flows. It should continue to produce results.

### Please tell us about your environment:

- [ ] Node-RED version: v0.20.6
- [ ] node.js version: v10.15.2
- [ ] npm version: v5.8.0
- [ ] Platform/OS: Raspberry Pi 10 Buster
- [ ] Browser: Vivaldi v2.5

### Cause
When the node closes it will close all the mcp-spi-adc objects that was created on startup for each channel. 

```
        node.on("close", function(done) {
            if (mcp3xxx.length !== 0) {
                var j=0;
                for (var i=0; i<8; i++) {
                    mcp3xxx[i].close(function() { j += 1; if (j === 8) {done()} });
                }
                mcp3xxx = [];
            }
            else { done(); }
        });
```

However, the code currently assumes that there will be 8 channels on closing. This works on an 8 channel chip, but when using a chip with less than 8 channels it will throw an unhandled exception because the last couple of mcp3xxx array members will be undefined.

dceejay:
Good spot - pushed a fix as version 0.2.1

PR #568, opened 2019-08-12
Close only the number of objects that was created

edreanernst:

## Proposed changes
This pull request addresses issue #567.

The close event of the node does not loop through the mcp3xxx array further than the amount of channels that was defined.
I have also added a try catch block to catch and log exceptions thrown by the close() method of the mcp-spi-adc object.

## Checklist


dceejay:
Sorry - saw the issue - didn't know you were going to raise a PR... so already fixed

PR #579, opened 2019-09-26
Add flow control to serialport

d-tamm:

## Proposed changes
Add flow control parameters for DTR and RTS. This is sometimes needed, e.g. with serial equipment which is powered abusing those pins.

## Checklist

### DIFF
--- a/io/serialport/25-serial.html
+++ b/io/serialport/25-serial.html
@@ -254,7 +254,7 @@
     <p>Provides configuration options for a serial port.</p>
     <p>The search button should return a list of available serial ports to choose from, or you
     can type in the location if known.</p>
-    <p>The DTR and RTS switches can be used to pull the flow control DTR and/or RTS pin high or low, e.g. in order to power devices via those pins. If not set, the pins will be low.</p>
+    <p>The DTR and RTS switches can be used to pull the flow control DTR and/or RTS pin high or low, e.g. in order to power devices via those pins.</p>
     <p>The node can optionally wait until it matches a pre-defined character.
         The data can then be split on a fixed character, after a timeout, or after a fixed number of characters.</p>
     <p>If using a character, it can be specified as either the character, the escaped shortcut (e.g. \n), or the hex code (e.g. 0x0d).</p>
--- a/io/serialport/25-serial.html
+++ b/io/serialport/25-serial.html
@@ -180,6 +180,22 @@
         </select>
         </td></tr></table>
     </div>
+    <div class="form-row">
+        <table width="100%">
+            <tr>
+                <td width="100px"></td>
+                <td width="150px">DTR <select type="text" id="node-config-input-dtr" style="width:100px;">
+                    <option value="high" data-i18n="serial.linestates.high"></option>
+                    <option value="low" data-i18n="serial.linestates.low"></option>
+                </select></td>
+                <td width="150px">RTS <select type="text" id="node-config-input-rts" style="width:100px;">
+                    <option value="high" data-i18n="serial.linestates.high"></option>
+                    <option value="low" data-i18n="serial.linestates.low"></option>
+                </select></td>
+                <td></td>
+            </tr>
+        </table>
+    </div>
     <div class="form-row" style="margin-top:16px; margin-bottom:0;">
         <label><i class="fa fa-sign-in"></i> <span data-i18n="serial.label.input"></span></label>
     </div>
@@ -238,6 +254,7 @@
     <p>Provides configuration options for a serial port.</p>
     <p>The search button should return a list of available serial ports to choose from, or you
     can type in the location if known.</p>
+    <p>The DTR and RTS switches can be used to pull the flow control DTR and/or RTS pin high or low, e.g. in order to power devices via those pins. If not set, the pins will be low.</p>
     <p>The node can optionally wait until it matches a pre-defined character.
         The data can then be split on a fixed character, after a timeout, or after a fixed number of characters.</p>
     <p>If using a character, it can be specified as either the character, the escaped shortcut (e.g. \n), or the hex code (e.g. 0x0d).</p>
@@ -254,6 +271,8 @@
             parity: {value:"none",required:true},
             stopbits: {value:1,required:true},
             waitfor: {value:""},
+            dtr: {value:"none"},
+            rts: {value:"none"},
             newline: {value:"\\n"},
             bin: {value:"false"},
             out: {value:"char"},
--- a/io/serialport/25-serial.html
+++ b/io/serialport/25-serial.html
@@ -154,6 +154,7 @@
             <td width="70px" data-i18n="serial.label.databits"></td>
             <td width="80px" data-i18n="serial.label.parity"></td>
             <td width="70px" data-i18n="serial.label.stopbits"></td>
+            <td></td>
         </tr><tr><td>&nbsp;</td>
         <td>
             <input type="text" id="node-config-input-serialbaud" style="width:92%">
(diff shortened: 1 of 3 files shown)


d-tamm:
I have tested the changes with a temperature data logger which is powered over the DTR and RTS lines, together with another logger which is externally powered (hence no settings for DTR and RTS). Both worked as expected.

dceejay:
Thanks for this. Before we accept it we need to understand a bit more about other possible options... A quick look would seem to indicate 

  a) that CTS and  DSR can also be controlled. 
  b) That actual flow control could potentially also be enabled (which would then preclude their use in this manner) and 
  c) Rather than set them here, may they need to be controlled via use of msg properties (eg to reset a device like an arduino). 

I think we need to consider some of these as there may be (or maybe not) a bigger picture we may need to address than just the specific options in this PR.

dceejay:
Thanks for this. Before we accept it we need to understand a bit more about other possible options... A quick look would seem to indicate 

  a) that CTS and  DSR can also be controlled. 
  b) That actual flow control could potentially also be enabled (which would then preclude their use in this manner) and 
  c) Rather than set them here, may they need to be controlled via use of msg properties (eg to reset a device like an arduino). 

I think we need to consider some of these as there may be (or maybe not) a bigger picture we may need to address than just the specific options in this PR.

d-tamm:
Thank you for looking at this. Frankly, I have quite limited knowledge about serial ports. I managed to hack this solution which was needed for me to be able to use nodered with my equipment - which for me was better than nothing. So I would need your (or other's) advice and cooperation for dealing with your questions.
a) Yes, we could add CTS And DSR, too, if you think that makes any sense. That should be easy, as they are controlled the same way as DTR and RTS (https://serialport.io/docs/api-stream).
b) I have no idea.
c) Sounds like an option to consider. Not sure though whether my knowledge and understanding are enough to implement that. You mean e.g. adding it at line 41 in the js file?

dceejay:
hi, sorry - we have been busy releasing v1.0 of Node-RED so not had a great chance to dig into this. My concerns still remain that these extra flags just fix your particular problem and don't address the wider capabilities of serial ports.

d-tamm:
Hi @dceejay . I think we should get to a conclusion in this issue, either by accepting or closing the PR. When I was working on this and googling for a solution to the need to power external devices via the flow control lines, I got several hits for for people/projects having the same problem, and solutions in different programming languages. My conclusion is therefore that this is a functionality which is needed by more people than just me.
I do not see either that implementing the flags as proposed would be a hinder in a possible later implementation of hardware flow control (your bullet b above), as long as we keep the possibility to have the flags unset, in which case they should not interfere with future flow control.
Please consider accepting the PR. Being able to contribute to OSS and add needed functionality is a major strength of OSS and was a reason for me to choose nodered. I would contribute and do the job of adding CTS and DSR in the same way as DTR and RTS are implemented right now.

dceejay:
OK - please add CTS and DSR to the same config line.
Also please add option `none` to each of the selects (with a label of " ") which should be the default (ie not set high or low)

io/serialport/25-serial.html
```diff
@@ -154,6 +154,7 @@ <h3>Outputs</h3>
             <td width="70px" data-i18n="serial.label.databits"></td>
             <td width="80px" data-i18n="serial.label.parity"></td>
             <td width="70px" data-i18n="serial.label.stopbits"></td>
+            <td></td>
```
dceejay: remove - not needed

io/serialport/25-serial.html
```diff
@@ -179,6 +180,22 @@ <h3>Outputs</h3>
         </select>
         </td></tr></table>
     </div>
+    <div class="form-row">
+        <table width="100%">
+            <tr>
+                <td width="100px"></td>
+                <td width="150px">DTR <select type="text" id="node-config-input-dtr" style="width:100px;">
+                    <option value="high" data-i18n="serial.linestates.high"></option>
```
dceejay: add option `none` - label " "

io/serialport/25-serial.html
```diff
@@ -179,6 +180,22 @@ <h3>Outputs</h3>
...
+                    <option value="low" data-i18n="serial.linestates.low"></option>
+                </select></td>
+                <td width="150px">RTS <select type="text" id="node-config-input-rts" style="width:100px;">
+                    <option value="high" data-i18n="serial.linestates.high"></option>
```
dceejay: add option "none"
as the default

io/serialport/25-serial.html
```diff
@@ -179,6 +180,22 @@ <h3>Outputs</h3>
...
+                    <option value="low" data-i18n="serial.linestates.low"></option>
+                </select></td>
+                <td></td>
```
dceejay: likewise add CTS and DSR

io/serialport/25-serial.html
```diff
@@ -253,6 +271,8 @@ <h3>Outputs</h3>
             parity: {value:"none",required:true},
             stopbits: {value:1,required:true},
             waitfor: {value:""},
+            dtr: {value:"none"},
+            rts: {value:"none"},
```
dceejay: and add cts, dsr

io/serialport/25-serial.html
```diff
@@ -237,6 +254,7 @@ <h3>Outputs</h3>
     <p>Provides configuration options for a serial port.</p>
     <p>The search button should return a list of available serial ports to choose from, or you
     can type in the location if known.</p>
+    <p>The DTR and RTS switches can be used to pull the flow control DTR and/or RTS pin high or low, e.g. in order to power devices via those pins.</p>
```
dceejay: and add cts, dsr

io/serialport/locales/en-US/25-serial.json
```diff
@@ -33,6 +33,10 @@
             "odd": "Odd",
             "space": "Space"
         },
+        "linestates": {
+            "high": "High",
+            "low": "Low"
+        },
```
dceejay: add `none` - (as " ")

io/serialport/25-serial.html
d-tamm: Yes it is, because all other cells on the row are fixed witdh. If we remove this flexible cell, the Baud rate cell will not align nicely.

io/serialport/25-serial.html
d-tamm: done

io/serialport/25-serial.html
d-tamm: done

io/serialport/25-serial.html
d-tamm: done

io/serialport/25-serial.html
d-tamm: done

io/serialport/25-serial.html
d-tamm: done

io/serialport/locales/en-US/25-serial.json
d-tamm: done (however, as empty string)

d-tamm:
This is how the settings dialog now looks like.
screenshot

I see that coverage decreases, and that that's supposed to be bad. I have no knowledge about what that means more exactly or why it decreases, but if there is anything I can do about it, please tell me.

io/serialport/25-serial.html
dceejay: or all 4 can be in their own table equally spaced.

dceejay:
No - looks all good now  :-)

dceejay:
now published as version 0.10.0

PR #653, opened 2020-05-20
On error sending pushover message throw a catchable exception

JsBergbau:
Throws an catchable exception if there is any error in sending the message. So you can react to it with a catch node

- [ x ] Bugfix (non-breaking change which fixes an issue)

## Proposed changes

If there is any error with sending a pushover message you can now react to it with a catch node and send it for example as an email. I'm using Pushover via curl and bash for some time now and sometimes there's an error submitting the messages to pushover. I then get an email as backup. So this change enables it in Node-RED to react when there is an error sending the message. 

## Checklist

- [ x ] I have read the [contribution guidelines](https://github.com/node-red/node-red-nodes/blob/master/CONTRIBUTING.md)

### DIFF
--- a/social/pushover/57-pushover.js
+++ b/social/pushover/57-pushover.js
@@ -91,16 +91,16 @@ module.exports = function(RED) {
                     node.error("[57-pushover.js] Error: attachment property must be a path to a local file or a Buffer containing an image");
                     return;
                 }
-                pushMessage(pushmsg);
+                pushMessage(pushmsg,msg);
             }
             else {
                 node.warn("Pushover credentials not set.");
             }
         });
 
-        function pushMessage(pushmsg) {
+        function pushMessage(pushmsg,msg) {
             pusher.send( pushmsg, function(err, response) {
-                if (err) { node.error(err,pushmsg); }
+                if (err) { node.error(err,msg); }
                 else {
                     try {
                         response = JSON.parse(response);
--- a/social/pushover/57-pushover.js
+++ b/social/pushover/57-pushover.js
@@ -100,7 +100,7 @@ module.exports = function(RED) {
 
         function pushMessage(pushmsg) {
             pusher.send( pushmsg, function(err, response) {
-                if (err) { node.error(err); }
+                if (err) { node.error(err,pushmsg); }
                 else {
                     try {
                         response = JSON.parse(response);


hardillb:
This change won't work, you need to pass the node's incoming `msg` object not the `pushmsg` that is passed to the `pushMessage()` function

social/pushover/57-pushover.js
```diff
@@ -100,7 +100,7 @@ module.exports = function(RED) {
 
         function pushMessage(pushmsg) {
             pusher.send( pushmsg, function(err, response) {
-                if (err) { node.error(err); }
+                if (err) { node.error(err,pushmsg); }
```
hardillb, line 103: `pushmsg` is not the incoming `msg` object.

hardillb:
Sorry, I should have said thank you for looking at this.

The `node.error()` function that takes 2 arguments expects the error and the input msg object that triggered the error.  Here you have passed the `pushmsg` object, this is a pushover specific object, not the input `msg`

dceejay:
Yes, the concept is good, just needs to be the correct msg.

JsBergbau:
Yeah youre right you could also pass the original msg object. And I considered and chose the `pushmsg`` because then you get the whole pushover message that was lost. 

You can argue what is better. Original message or pushover message or the message you passed to pushover node?

The `pushmsg` includes much more details like retry, expires, priority, URL and so on. A lot of stuff you directly configure in the pushover node. If you pass the original message object, this information is lost.
grafik
Output at the top with catch node catching the error, output at the bottom with disabled catch node. So no change in behaviour.

I prefer to throw the `pushmsg` instead of the original message because then you have much more possibilities to react, like sending to a specific device. 

grafik

To do that with the original msg object you have redundant places where you specify where your message should go and thus is more error prone.

 However if it is required to throw the original message to get it integrated, I'll rewrite to do that.

hardillb:
It's not a case of arguing which is better, it's what the API says you have to pass.

Please update it to return the original input msg object.

JsBergbau:
Ok, I've updated. Would it be according to the API to set `msg.error` to `pushmsg` then it is move to `msg._error` and could be accessed in the exception handling code. If so, I would add a new commit.

https://nodered.org/docs/user-guide/handling-errors 

> If the message already had a msg.error property when the node reported the error, that property will be moved to msg._error.

dceejay:
Not really - the error node adds msg.error.
What extra info that the pushover node adds do you really need that can't be recreated from msg by whatever error handling you have ?

JsBergbau:
> Not really - the error node adds msg.error.

Just to make sure that there is no misunderstanding: If we set `msg.error` to `pushmsg` and then throw the exception, than the `pushmsg` can be accessed via msg._error. But if this behaviour is also not API conform then I recreate all the details by another way.

dceejay:
the intention of _error is a push stack created by other error nodes - so that a tight loop of a repeated errors could be detected.

JsBergbau:
Ok then I'll leave the code that way and it can be integrated.

dceejay:
Great many thanks for raising and fixing this.
published as version 0.0.19

PR #675, opened 2020-08-26
[serialport] Allow baudrate change at run time

Orfait:

## Proposed changes

Allow baudrate modification during run time of the flow.
https://discourse.nodered.org/t/node-red-node-serialport-change-baudrate-on-the-fly/4313

## Checklist


io/serialport/25-serial.js
```diff
@@ -249,6 +277,7 @@ module.exports = function(RED) {
                             return payload;
                         },
                         write: function(m,cb) { this.serial.write(m,cb); },
+			update: function(m,cb) { this.serial.update(m,cb); },
```
dceejay, line 280: alignment

io/serialport/25-serial.js
```diff
@@ -121,6 +135,20 @@ module.exports = function(RED) {
             node.port = serialPool.get(this.serialConfig);
             // Serial Out
             node.on("input",function(msg) {
+                if (msg.hasOwnProperty("baudrate")) {
+					var baud = parseInt(msg.baudrate);
+					if (isNaN(baud)) {
+						var errmsg = err.toString().replace("Serialport","Serialport "+node.port.serial.path);
+						node.error(errmsg,"Cannot parse baudrate property");
```
dceejay, line 142: as per above errmsg should be the text, 2nd param should just be msg.
It should also be translatable - the text should be in the locale file - eg see line 61 etc.

io/serialport/25-serial.js
```diff
@@ -41,6 +41,20 @@ module.exports = function(RED) {
             node.port = serialPool.get(this.serialConfig);
 
             node.on("input",function(msg) {
+				if (msg.hasOwnProperty("baudrate")) {
+					var baud = parseInt(msg.baudrate);
+					if (isNaN(baud)) {
+						var errmsg = err.toString().replace("Serialport","Serialport "+node.port.serial.path);
+						node.error(errmsg,"Cannot parse baudrate property");
```
dceejay, line 48: the errmsg should be the text you have, and the second parameter should just be the incoming msg.
They should also be translatable - the text should be in the locale file - eg see line 61 etc.

io/serialport/25-serial.js
Orfait: Changed like line 161.

io/serialport/25-serial.js
Orfait: Changed to be like line 63

io/serialport/25-serial.js
```diff
@@ -41,6 +41,20 @@ module.exports = function(RED) {
             node.port = serialPool.get(this.serialConfig);
 
             node.on("input",function(msg) {
+				if (msg.hasOwnProperty("baudrate")) {
+					var baud = parseInt(msg.baudrate);
+					if (isNaN(baud)) {
+						var errmsg = err.toString().replace("Serialport","Serialport "+node.port.serial.path);
```
dceejay, line 47: There is no err at this point - you should just set it to the string you want. But the string needs to be the translated one - so something like `node.error(RED._("serial.errors.badbaudrate"),msg);`
and then create a badbaudrate entry in the US locale file.

Orfait: I did it, but I must admit it is over my actual knowledge.

dceejay:
Thanks

Orfait:
Hi @dceejay 
I forgot to change the version on the node (actually 0.10.3). Is it needed ?
Maybe this is done when making a release ?

dceejay:
no need - yes I will change it... - I also want to update the underlying serialport lib so have included that and am testing before release.

PR #676, opened 2020-08-26
Enable the use of environment variables to define GPIO pin number

pmacostapdi:
The "pin" input field is exposed so that the pin number may be typed in
directly (as opposed to selecting it from the pin table), and also
environment variables may be used to define the pin number. This, in turn,
enables the use of the GPIO pins a subflow template such that multiple
sublfow instances of that template can be running in the same Raspberry Pi
with different pin numbers.

## Checklist

### DIFF
--- a/hardware/PiGpio/36-rpi-gpio.html
+++ b/hardware/PiGpio/36-rpi-gpio.html
@@ -70,9 +70,7 @@
 <script type="text/html" data-template-name="rpi-gpio in">
 
     <div class="form-row" style="min-width: 540px">
-        <label>&nbsp;&nbsp;&nbsp;&nbsp;<span data-i18n="rpi-gpio.pinname"></span></label>
-        <input type="text" id="node-input-pin">
-        <div><p></p><div><label></label>
+		<label><i class="fa fa-circle"></i> <span data-i18n="rpi-gpio.pinname"></span></label>
         <div class="rpi-gpio-pinTable">
             <div class="pinTableBody" id="pinform">
                 <div class="pinTableRow">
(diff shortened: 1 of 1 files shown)


hardware/PiGpio/36-rpi-gpio.html
```diff
@@ -70,8 +70,9 @@
 <script type="text/html" data-template-name="rpi-gpio in">
 
     <div class="form-row" style="min-width: 540px">
-        <label><i class="fa fa-circle"></i> <span data-i18n="rpi-gpio.pinname"></span></label>
-        <input type="text" id="node-input-pin" style="display:none;">
+        <label>&nbsp;&nbsp;&nbsp;&nbsp;<span data-i18n="rpi-gpio.pinname"></span></label>
```
dceejay, line 73: I think I'd prefer the new field to be arranged below the existing selector so that by the time someone scrolls down it may already match the pin they have selected. (and leave the fa-icon)

hardware/PiGpio/36-rpi-gpio.html
```diff
@@ -255,8 +276,9 @@
 
 <script type="text/html" data-template-name="rpi-gpio out">
     <div class="form-row" style="min-width: 540px">
-        <label><i class="fa fa-circle"></i> <span data-i18n="rpi-gpio.pinname"></span></label>
-        <input type="text" id="node-input-pin" style="display:none;">
+        <label>&nbsp;&nbsp;&nbsp;&nbsp;<span data-i18n="rpi-gpio.pinname"></span></label>
```
dceejay, line 279: I think I'd prefer the new field to be arranged below the existing selector so that by the time someone scrolls down it may already match the pin they have selected. (and leave the fa-icon)

dceejay (changes requested):
thanks for the PR.

hardware/PiGpio/36-rpi-gpio.html
pmacostapdi: Same comments as above.

hardware/PiGpio/36-rpi-gpio.html
pmacostapdi: I incorporated your suggestions in my second commit. I also closed validation escapes when directly typing the pin number.

dceejay (approved):

dceejay:
Looks good to me. thanks again.

Issue #673, opened 2020-08-21
Pushbullet input node tries to reconnect too fast


elmicha:
### Which node are you reporting an issue on?

node-red-node-pushbullet 0.0.14

### What are the steps to reproduce?

I'm not sure. Maybe the Pushbullet server had a problem, maybe my network or the internet to the pushbullet server had a problem.

### What happens?

The `pushbullet in` node had a problem connecting to the Pushbullet server:

```
Aug 21 15:28:54 rpi Node-RED[566]: 21 Aug 15:28:54 - [error] [pushbullet in:43cf06f2.df7d58] Error: Unexpected server response: 502
Aug 21 15:28:58 rpi Node-RED[566]: 21 Aug 15:28:58 - [error] [pushbullet in:43cf06f2.df7d58] Error: Unexpected server response: 502
Aug 21 15:29:18 rpi Node-RED[566]: 21 Aug 15:29:18 - [error] [pushbullet in:43cf06f2.df7d58] Error: Unexpected server response: 502
```

That's ok, it can happen, and trying it again after a few seconds also is ok. But after a few minutes, the pushbullet in node becomes a bit impatient, and the Pushbullet server answers with HTTP error 429, too many requests:

```
Aug 21 15:41:06 rpi Node-RED[566]: 21 Aug 15:41:06 - [error] [pushbullet in:43cf06f2.df7d58] Error: Unexpected server response: 502
Aug 21 15:41:08 rpi Node-RED[566]: 21 Aug 15:41:08 - [error] [pushbullet in:43cf06f2.df7d58] Error: Unexpected server response: 502
Aug 21 15:41:08 rpi Node-RED[566]: 21 Aug 15:41:08 - [error] [pushbullet in:43cf06f2.df7d58] Error: read ECONNRESET
Aug 21 15:41:08 rpi Node-RED[566]: 21 Aug 15:41:08 - [error] [pushbullet in:43cf06f2.df7d58] Error: read ECONNRESET
Aug 21 15:41:08 rpi Node-RED[566]: 21 Aug 15:41:08 - [error] [pushbullet in:43cf06f2.df7d58] Error: read ECONNRESET
Aug 21 15:41:08 rpi Node-RED[566]: 21 Aug 15:41:08 - [error] [pushbullet in:43cf06f2.df7d58] Error: read ECONNRESET
[...]

PR #701, opened 2020-10-13
Attempt to mitigate stream server connect feedback loop

guzba:
Pushbullet dev here. Around 36 hours ago we started receiving around 750 additional requests per second to connect to our stream server. I blocked around 50 - 100 IP addresses to mitigate the spam. This spam traffic has continued unchanged since then.

I don't know for sure what caused this, but one user came forward reporting issues with their Node-Red setup after I banned the IPs (status code is my banned IP status code which is pretty compelling evidence). I can't be sure what is causing the issue since no User-Agent is included in the requests, but it for sure is at least partly Node-Red API users. I'd like to work on mitigating this since mitigating the spam traffic costs money.

My first theory here is in this PR. I am not aware of any guarantees that you'll never get multiple 'error' events, so it makes sense to clear the timeout before setting a new one to reconnect. Even if it is just defensive, it may not have much cost? If you do get multiple error events without this change, it is possible to trigger many connect() requests which can error out and then trigger many more connect() requests with lots of timeouts, which is not good.

Another thing to consider is that this library's dependency for PB (https://github.com/alexwhitman/node-pushbullet-api) already has code to reconnect if a websocket connection is lost without having been properly closed. I think an even better solution is verifying this works and then deleting the reconnect logic from here.

I'm not an expert on Node-Red though so I am only trying to offer suggestions to ensure everyone is a good citizen when using our API :) Thanks!

## Checklist

### DIFF
--- a/social/pushbullet/57-pushbullet.js
+++ b/social/pushbullet/57-pushbullet.js
@@ -125,6 +125,7 @@ module.exports = function(RED) {
             stream.on('close', function() {
                 self.emitter.emit('stream_disconnected');
                 if (!closing) {
+                    if (tout) { clearTimeout(tout); }
                     tout = setTimeout(function() {
                         stream.connect();
                     }, 15000);
@@ -133,6 +134,7 @@ module.exports = function(RED) {
             stream.on('error', function(err) {
                 self.emitter.emit('stream_error', err);
                 if (!closing) {
+                    if (tout) { clearTimeout(tout); }
                     tout = setTimeout(function() {
                         stream.connect();
                     }, 15000);


guzba:
Related: https://github.com/node-red/node-red-nodes/issues/673

knolleary:
Hi @guzba - thanks for this. My apologies for the excess traffic this node has caused you - it isn't a node we've paid much attention to as it has generally just worked, but I can see how this fix would prevent one possible cause of reconnect storms.

I'll merge this and publish an update now. We can see if it improves matters as the community upgrade to the new version.

guzba:
@knolleary No worries at all, I totally get how it goes with programming. That's just how it is. Hope I didn't come off as upset or anything like that! Just trying to help since I don't think this is obvious without seeing from my side.

Amazingly quick response here. I'll stay in the loop here to see if I can confirm some positive improvement.

Issue #679, opened 2020-09-05
Multiple SNMP Triggers Get Grouped Together


skeletorsue:
### Which node are you reporting an issue on?
SNMP

### What are the steps to reproduce?
If you send multiple concurrent requests to snmp.subtree the responses will get jumbled up with each other. I've set up a stripped down test set up with docker-compose and a couple of mock snmp servers for testing. I will attach a zip file containing the the docker-compose set up (docker-compose.yaml) and custom walk file that the mock snmp servers use to give their responses (custom.walk). I will also provide an export of the flows.  One is to show that initiating 3 requests at the same time results in combined results and the other shows that if you add delay and stagger the requests the debug output is as expected.

Flow Exports
[...]

PR #705, opened 2020-10-24
Added support for V3

ASH3xC0d3:
Added support for version 3.
All the existing issues with v1 and v2c are also with v3 e.g.
https://github.com/node-red/node-red-nodes/issues/679
snmp walker will not return more than 1000 objects, all following objects cannot be accessed
Incorrect encryption (privacy) passphrase gives a timeout and not an error (protocol limitation?)
Incorrect community string gives a timeout and not an error (protocol limitation?)
Changes to the node are made with following details:
Node-red v1.2.2
Node.js v12.6.0
Windows_NT 10.0.19841 x64 LE

## Checklist

### DIFF
--- a/io/snmp/snmp.html
+++ b/io/snmp/snmp.html
@@ -4,18 +4,55 @@
         <input type="text" id="node-input-host" placeholder="ip address(:optional port)">
     </div>
     <div class="form-row">
-        <label for="node-input-community"><i class="fa fa-user"></i> Community</label>
-        <input type="text" id="node-input-community" placeholder="public">
-    </div>
-    <div class="form-row">
         <label for="node-input-version"><i class="fa fa-bookmark"></i> Version</label>
         <select type="text" id="node-input-version" style="width:150px;">
-            <option value="1">v1</option>
-            <option value="2c">v2c</option>
+            <option value="v1">v1</option>
+            <option value="v2c">v2c</option>
+           <option value="v3">v3</option>
         </select>
         <span style="margin-left:50px;">Timeout</span>
         <input type="text" id="node-input-timeout" placeholder="secs" style="width:50px; direction:rtl; vertical-align:baseline;">&nbsp;S
     </div>
+    <div class="form-row">
+        <label for="node-input-community"><i class="fa fa-user"></i> Community</label>
+        <input type="text" id="node-input-community" placeholder="public">
+    </div>
+	<!-- Following Data is used for V3 Only -->
+	<div class="form-row">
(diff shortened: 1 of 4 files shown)


io/snmp/snmp.html
```diff
@@ -3,19 +3,56 @@
...
+            <option value="SHA">SHA</option>
+        </select>
+        <span style="margin-left:50px;">Priv.Prot.</span>
+        <select type="text" id="node-input-privprot" style="width:150px;">
+            <option value="None">None</option>
+            <option value="DES">DES</option>
+            <option value="AES">AES</option>
+        </select>
+    </div>
+    <div class="form-row">
```
dceejay, line 47: Likewise - do these need to be shown if "None" is selected above ?

io/snmp/snmp.html
```diff
@@ -3,19 +3,56 @@
...
+        <label for="node-input-authkey"><i class="fa fa-user"></i> Auth.Key</label>
+        <input type="password" id="node-input-authkey" placeholder="Authentication key">
+    </div>
+    <div class="form-row">
```
dceejay, line 51: Same again

io/snmp/snmp.html
```diff
@@ -42,10 +82,16 @@
...
             host: { value: "127.0.0.1" },
+            version: { value: "v1", required: true },
+            timeout: { value: 5 },
             community: { value: "public" },
-            version: { value: "1", required: true },
+            username: { value: "" },
+            auth: { value: "noAuthNoPriv", required: true },
+            authprot: { value: "None", required: true },
+            privprot: { value: "None", required: true },
+            authkey: { value: "" },
```
dceejay, line 92: Any credentials need to be handled as credentials and not general variables - see https://nodered.org/docs/creating-nodes/credentials (ie add a credential section like the defaults one but for the items that should be secret.)

dceejay (changes requested):
So yes - two main things - 
1) is that we should hide any fields from the user that aren't necessary - so only show v3 relevant fields when v3 is selected. and keys when encryption is selected etc... A simple example to follow is in the smooth node 17-smooth.html line 35 and then 91, 96 . 
2) All credentials MUST be stored as credentials for security - so need to be handles as per docs - https://nodered.org/docs/creating-nodes/credentials - or for an example see the mysql node 68-mysql.html line 48 and 68-mysql.js line 39 and 97.

Steve-Mcl:
@ASH3xC0d3 your changes have been incorporated into an updated PR #904 and merged into master

Thanks for the PR 👍

PR #707, opened 2020-10-26
Update node-red-random to add option to dynamically pass in 'To' and 'From'

juggledad:
- [ x ] New feature (non-breaking change which adds functionality)

## Proposed changes

This update allows the user to set the 'To' and 'From' values in an incoming msg as 'msg.to' and 'msg.from'. In my use case, I wanted to pick a random file from a number of folders each with a different number of files. Without this change, I would need to have one random node for each folder and would have to change the hard coded values if a file was added or remove from the folder.

In addition, I discovered a but where if 'From' was greater than 'To' you would only get numbers between them. So if 'From' was 3 and 'To' was 1 you wold only receive 2.
 
## Checklist

- [ x] I have read the [contribution guidelines](https://github.com/node-red/node-red-nodes/blob/master/CONTRIBUTING.md)
- [ x] For non-bugfix PRs, I have discussed this change on the forum/slack team.

### DIFF
--- a/function/random/random.html
+++ b/function/random/random.html
@@ -31,8 +31,8 @@
         color:"#E2D96E",
         defaults: {
             name: {value:""},
-            low: {value:""},
-            high: {value:""},
+            low:  {value:"",validate:function(v) { return  !isNaN(v) || v.length === 0;} },
+            high: {value:"",validate:function(v) { return  !isNaN(v) || v.length === 0;} },
             inte: {value:"true"},
             property: {value:"payload",required:true}
         },
--- a/function/random/random.html
+++ b/function/random/random.html
@@ -31,8 +31,8 @@
         color:"#E2D96E",
         defaults: {
             name: {value:""},
-            low: {value:"1"},
-            high: {value:"10"},
+            low: {value:""},
+            high: {value:""},
             inte: {value:"true"},
             property: {value:"payload",required:true}
         },
--- a/function/random/random.html
+++ b/function/random/random.html
@@ -31,8 +31,8 @@
         color:"#E2D96E",
         defaults: {
             name: {value:""},
-            low:  {value:"",validate:function(v) { return  !isNaN(v) || v.length === 0;} },
-            high: {value:"",validate:function(v) { return  !isNaN(v) || v.length === 0;} },
+            low:  {value: 1,validate:function(v) { return  !isNaN(v) || v.length === 0;} },
+            high: {value: 10,validate:function(v) { return  !isNaN(v) || v.length === 0;} },
             inte: {value:"true"},
             property: {value:"payload",required:true}
         },
--- a/function/random/random.js
+++ b/function/random/random.js
@@ -3,13 +3,44 @@ module.exports = function(RED) {
     "use strict";
     function RandomNode(n) {
         RED.nodes.createNode(this,n);
-        this.low = Number(n.low || 1);
-        this.high = Number(n.high || 10);
+
+        this.low  = Number(n.low) || ""; 
+        this.high = Number(n.high) || "";
         this.inte = n.inte || false;
         this.property = n.property||"payload";
         var node = this;
+        
+      
         this.on("input", function(msg) {
+
+            if (node.low === "") {
+                node.low = 1
+                if ('from' in msg) {
+                    if ( (typeof msg.from === 'number') 
+                    || ( (typeof msg.from === 'string') && (!isNaN(Number(msg.from)) ) )) {
+                        node.low = Number(msg.from)
+                    }
+                } 
+            }
+    
+            if (node.high === "") {
+                node.high = 10 
+                if ('to' in msg) {
+                    if ( (typeof msg.to === 'number') 
+                    || ( (typeof msg.to === 'string') && (!isNaN(Number(msg.to)) ) )) {
+                        node.high = Number(msg.to)
+                    }
+                } 
+            }
+            
+            // flip the values if low > high
             var value;
+            if (node.low > node.high) {  
+                value = node.low
+                node.low = node.high
+                node.high = value
+            }
+
             if (node.inte == "true" || node.inte === true) {
                 value = Math.round(Math.random() * (node.high - node.low + 1) + node.low - 0.5);
             }
--- a/function/random/random.js
+++ b/function/random/random.js
@@ -12,7 +12,6 @@ module.exports = function(RED) {
         var tmp = {};
 
         this.on("input", function(msg) {
-          this.status({}); // blank out the status on a deploy
 
           tmp.low = 1                 // set this as the default low value
           tmp.low_e = ""
@@ -69,7 +68,6 @@ module.exports = function(RED) {
              }
             
              RED.util.setMessageProperty(msg,node.property,value);
-             this.status({fill:"grey",shape:"dot",text:"random: from " + tmp.low + " to " + tmp.high});
              node.send(msg);
           }
         });
--- a/function/random/random.js
+++ b/function/random/random.js
@@ -15,31 +15,34 @@ module.exports = function(RED) {
           this.status({}); // blank out the status on a deploy
 
           tmp.low = 1                 // set this as the default low value
+          tmp.low_e = ""
           if (node.low) {             // if the the node has a value use it
               tmp.low = node.low
           } else if ('from' in msg) { // else see if a 'from' is in the msg
               if (Number(msg.from)) { // if it is, and is a number, use it
                   tmp.low = Number(msg.from);
               } else {                // otherwise setup NaN error
-                  tmp.low = NaN
+                  tmp.low = NaN;
+                  tmp.low_e = " From: " + msg.from; // setup to show bad incoming msg.from
               }
           } 
               
           tmp.high = 10               // set this as the default high value  
-          if (node.high) {            // if the the node has a value use it
+          tmp.high_e = "";
+          if (node.high) {            // if the the node has a value use it 
               tmp.high = node.high
           } else if ('to' in msg) {   // else see if a 'to' is in the msg
               if (Number(msg.to)) {   // if it is, and is a number, use it
                   tmp.high = Number(msg.to);
               } else {                // otherwise setup NaN error 
                   tmp.high = NaN
+                  tmp.high_e = " To: " + msg.to // setup to show bad incoming msg.to
               }
           } 
                            
-          // if tmp.low or high are not numbers, send an error msg
+          // if tmp.low or high are not numbers, send an error msg with bad values
           if ( (isNaN(tmp.low)) || (isNaN(tmp.high)) ) {
-              this.status({fill:"red",shape:"dot",text:"random: from " + Number(tmp.low) + " to " + Number(tmp.high)})
-              this.error("Random: one of the input values is not a number")
+              this.error("Random: one of the input values is not a number. " + tmp.low_e + tmp.high_e);
           } else {
           // at this point we have valid values so now to generate the random number! 
 
--- a/function/random/random.js
+++ b/function/random/random.js
@@ -16,8 +16,7 @@ module.exports = function(RED) {
             if (node.low === "") {
                 node.low = 1
                 if ('from' in msg) {
-                    if ( (typeof msg.from === 'number') 
-                    || ( (typeof msg.from === 'string') && (!isNaN(Number(msg.from)) ) )) {
+                    if ( (typeof msg.from === 'number') || ( (typeof msg.from === 'string') && (!isNaN(Number(msg.from)) ) )) {
                         node.low = Number(msg.from)
                     }
                 } 
@@ -26,8 +25,7 @@ module.exports = function(RED) {
             if (node.high === "") {
                 node.high = 10 
                 if ('to' in msg) {
-                    if ( (typeof msg.to === 'number') 
-                    || ( (typeof msg.to === 'string') && (!isNaN(Number(msg.to)) ) )) {
+                    if ( (typeof msg.to === 'number') || ( (typeof msg.to === 'string') && (!isNaN(Number(msg.to)) ) )) {
                         node.high = Number(msg.to)
                     }
                 } 
--- a/function/random/random.js
+++ b/function/random/random.js
@@ -9,31 +9,32 @@ module.exports = function(RED) {
         this.inte = n.inte || false;
         this.property = n.property||"payload";
         var node = this;
-        var tmp = {};  
+        var tmp = {};
 
         this.on("input", function(msg) {
 
-           if (node.low === "") {
-               tmp.low = 1;
-               if ('from' in msg) {
-                 tmp.low = msg.from} 
+           if (node.low) {            // if the the node has a value use it
+               tmp.low = node.low
+           } else {                   // if 'from' in the msg, use it or default to 1
+               tmp.low = ('from' in msg) ? msg.from : 1;
            }
-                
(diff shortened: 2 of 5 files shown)


juggledad:
In random.js I had coded two IF statements eaach spanning two lines to make them readable. I have corrected this and made tehm each one line.

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
...
         this.on("input", function(msg) {
+ 
+           if (n.low !== "") {this.low  = parseFloat(n.low)}
+              else {this.low = n.low}; 
+           if (n.high !== "") {this.high  = parseFloat(n.high)}
+              else {this.high = n.high}; 
+      
+           if (node.low === "") {
+               node.low = 1;
+               if ('from' in msg) {node.low = msg.from}
```
dceejay, line 23: we can't just change node.low as that would then become a permanant change next time it was called. must use a a new variable for this.

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
...
+              else {this.high = n.high}; 
+      
+           if (node.low === "") {
+               node.low = 1;
+               if ('from' in msg) {node.low = msg.from} 
+           }
+                
+           if (node.high === "") {
+               node.high = 10; 
+               if ('to' in msg) {node.high = msg.to}
```
dceejay, line 28: likewise can't just change node.high

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
...
+               if ('from' in msg) {node.low = msg.from} 
+           }
+                
+           if (node.high === "") {
+               node.high = 10; 
+               if ('to' in msg) {node.high = msg.to} 
+           }
+
+           // if returning an interger, do a parseInt() on the low and high values
+           if ( (node.inte == "true") || (node.inte === true) ) {
```
dceejay, line 32: not sure why we need to force the limits to be integers. Just because we set the output to be integers, the limits could still be floats. What is wrong with asking for an integer between 1.2 and 8.7 ?

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
...
-                value = Math.random() * (node.high - node.low) + node.low;
-            }
-            RED.util.setMessageProperty(msg,node.property,value);
-            node.send(msg);
+           RED.util.setMessageProperty(msg,node.property,value);
+
+           // output a status under the node
+           var color = "grey";
+           if ( (isNaN(node.low)) || (isNaN(node.high)) ) {
+              this.status({fill:"red",shape:"dot",text:"random: from " + node.low + " to " + node.high})
```
dceejay, line 58: Not sure we need status for this sort of thing - just increases traffic to front end. Yes we need the node.error

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
...
+              this.error({node:"random",text:"one of the input values is not a number"})
+           } else {
+              this.status({fill:"grey",shape:"dot",text:"random: from " + node.low + " to " + node.high + " value=" + value});
```
dceejay, line 61: as above

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
     function RandomNode(n) {
         RED.nodes.createNode(this,n);
 
-        this.low  = Number(n.low) || ""; 
-        this.high = Number(n.high) || "";
+        this.status({}); // blank status on a deploy
```
dceejay, line 7: Not sure we need status for this sort of thing. Maybe use hoverover on output pin.

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
...
         this.inte = n.inte || false;
         this.property = n.property||"payload";
         var node = this;
-        
-      
+
+
         this.on("input", function(msg) {
+ 
+           if (n.low !== "") {this.low  = parseFloat(n.low)}
```
dceejay, line 16: This can be done outside of the msg on input as it only needs to be set once.

function/random/random.js
```diff
@@ -4,49 +4,63 @@ module.exports = function(RED) {
...
+              else {this.low = n.low}; 
+           if (n.high !== "") {this.high  = parseFloat(n.high)}
```
dceejay, line 18: likewise

dceejay (changes requested):

function/random/random.html
```diff
@@ -31,8 +31,8 @@
         color:"#E2D96E",
         defaults: {
             name: {value:""},
-            low: {value:"1"},
-            high: {value:"10"},
+            low:  {value:"",validate:function(v) { return  !isNaN(v) || v.length === 0;} },
```
dceejay, line 34: why can't we still default it here ? The user can always delete it if the wish.

function/random/random.html
```diff
@@ -31,8 +31,8 @@
...
+            high: {value:"",validate:function(v) { return  !isNaN(v) || v.length === 0;} },
```
dceejay, line 35: same here for high limit default.

function/random/random.js
juggledad: This was fixed by using a new variable

function/random/random.js
juggledad: same fix

function/random/random.js
juggledad: status removed - I found it helpful in the editor when sending different from/to's to the node

dceejay:
Defaults are Ok so it works as-is  - but a simple test setting min seems to fail .
Plus need some docs in the info bar to explain the new possible inputs.

```
[{"id":"e28f0688.605608","type":"ui_button","z":"e9b4dddd.70115","name":"","group":"463fb27.9c2064c","order":5,"width":0,"height":0,"passthru":false,"label":"button","tooltip":"","color":"","bgcolor":"","icon":"","payload":"5","payloadType":"str","topic":"","x":170,"y":1020,"wires":[["27e2a825.9149d8"]]},{"id":"a294fa0f.4c7228","type":"ui_gauge","z":"e9b4dddd.70115","name":"","group":"463fb27.9c2064c","order":6,"width":0,"height":0,"gtype":"gage","title":"gauge","label":"units","format":"{{value | number:2}}","min":0,"max":10,"colors":["#00b500","#e6e600","#ca3838"],"seg1":"","seg2":"","x":550,"y":1060,"wires":[]},{"id":"90c46d93.712fd","type":"random","z":"e9b4dddd.70115","name":"","low":"","high":"10","inte":"false","property":"payload","x":390,"y":1020,"wires":[["a294fa0f.4c7228","7533fdba.bbdb14"]]},{"id":"27e2a825.9149d8","type":"change","z":"e9b4dddd.70115","name":"","rules":[{"t":"set","p":"min","pt":"msg","to":"payload","tot":"msg"}],"action":"","property":"","from":"","to":"","reg":false,"x":270,"y":1080,"wires":[["90c46d93.712fd"]]},{"id":"7533fdba.bbdb14","type":"ui_chart","z":"e9b4dddd.70115","name":"","group":"463fb27.9c2064c","order":5,"width":0,"height":0,"label":"chart","chartType":"line","legend":"false","xformat":"HH:mm:ss","interpolate":"linear","nodata":"","dot":false,"ymin":"","ymax":"","removeOlder":1,"removeOlderPoints":"","removeOlderUnit":"60","cutout":0,"useOneColor":false,"useUTC":false,"colors":["#1f77b4","#aec7e8","#ff7f0e","#2ca02c","#98df8a","#d62728","#ff9896","#9467bd","#c5b0d5"],"outputs":1,"x":550,"y":1020,"wires":[[]]},{"id":"463fb27.9c2064c","type":"ui_group","name":"test","tab":"cc2c465a.11e6d","order":1,"disp":true,"width":"8","collapse":false},{"id":"cc2c465a.11e6d","type":"ui_tab","name":"test","icon":"dashboard","order":1,"disabled":false,"hidden":false}]
```

dceejay:
Much better - but is still doesn't warn if the from or to properties are missing

```
[{"id":"61e2e911.df36d8","type":"random","z":"e9b4dddd.70115","name":"","low":"","high":"10","inte":"true","property":"payload","x":500,"y":1140,"wires":[["fba2b43a.1a4f88"]]},{"id":"fba2b43a.1a4f88","type":"debug","z":"e9b4dddd.70115","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":650,"y":1140,"wires":[]},{"id":"59267f22.8d495","type":"random","z":"e9b4dddd.70115","name":"","low":"1","high":"","inte":"true","property":"payload","x":500,"y":1200,"wires":[["ec73089e.02b298"]]},{"id":"ec73089e.02b298","type":"debug","z":"e9b4dddd.70115","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":650,"y":1200,"wires":[]},{"id":"850b6261.37933","type":"random","z":"e9b4dddd.70115","name":"","low":"","high":"","inte":"true","property":"payload","x":500,"y":1260,"wires":[["b73537bc.e39e78"]]},{"id":"b73537bc.e39e78","type":"debug","z":"e9b4dddd.70115","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":650,"y":1260,"wires":[]},{"id":"aff2373b.e90568","type":"inject","z":"e9b4dddd.70115","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"","payloadType":"date","x":250,"y":1200,"wires":[["61e2e911.df36d8","59267f22.8d495","850b6261.37933"]]}]
```

juggledad:
When you say there is no warning if the from or to is missing, I assume (correct me if I am wrong) you mean msg.from and msg.to - correct?

Why should there be a warning? If there is something in the node property for 'to' and I only send in msg.from it works. Even if there is nothing for 'to' in the node, 'to' defaults to 10 and you get a result.  In fact, I think the node, when added to a flow, should have 'from' and 'to' NOT filled in. If nothing comes in, they will default to 1 and 10. If the node (when added to tehh flow) fills from and to in by default and I add a inject sending msg.from and/or msg.to, those will be ignored because the node will have values already in it. So the user has to remember everytime they add a random node in, if they want to override it, they have to blank the values out (this has already bitten me!)

And what about the case where I want to get a random number from 1 to N  and N (msg.to) will vary based on something else but the from will always be 1? (In my case, I want a camera to recognize a face and based on the face, get the count of files in a folder associated with that face and pick a random file from that file. Then number of files in the folder will vary from face to face.) 

But I will abide to your decision.

function/random/random.js
```diff
@@ -3,21 +3,60 @@ module.exports = function(RED) {
...
+               tmp.high = ('to' in msg) ? msg.to : 10;
+           }
+           
+          // if tmp.low or high are not numbers, send an error msg
+          if ( (isNaN(tmp.low)) || (isNaN(tmp.high)) ) {
+              this.error({node:"random",text:"one of the input values is not a number"})
+          } 
+          else {
+             // force tmp.high/low to Numbers since they may be a string
+             tmp.low = Number(tmp.low)
```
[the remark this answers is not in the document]
juggledad: Does it really matter if if it is done before the test to see if they are numbers? 
         // if tmp.low or high are not numbers, send an error msg
           if ( (isNaN(tmp.low)) || (isNaN(tmp.high)) ) {
               this.error({node:"random",text:"one of the input values is not a number"})
           } 
           else {
              // force tmp.high/low to Numbers since they may be a string
              tmp.low = Number(tmp.low)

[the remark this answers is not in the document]
dceejay: probably not but this is javascript so there may be some weird edge cases where isNaN says is an OK number then Number fails to convert it.

dceejay:
Sorry - I keep confusing things.   
Right now (existing node) - drag it on 
no config - give a number from 1-10. 
If I open and save config (but don't change anything) - I still get 1 - 10.

So both of those behaviours should remain - even if I send in msg.from and/or msg.to.

New version
If I drag on - no config - then node.from and node.to are undefined - and we should still get 1-10.
if I edit config and save (no changes) - node.from and node.to are now 1 and 10 - so should get from 1 to 10.
If I edit config and blank out from then node.from === ""  (not undefined) - so now msg.from MUST set it or produce warning/error.
If I edit config and blank out to then node.to === ""  (not undefined) - so now msg.to MUST set it or produce warning/error.

juggledad:
Dave, there is one more case with the original version
case 1 - add node, don't edit it - you get 1-10
case 2 - edit node but don't change anything - you get 1-10
case 3 - edit node and remove 'From' and 'To' so they are blank - you get 1 - 10
With the original version, if 'From' or 'To' is omitted the default value 1 or 10 is used. So if 'From' is 5 and 'To' is blank, the result will be from 5-10.

That's what I've tried to replicate in the new version. If 'From' and/or 'To' is defined in the node that takes precedence. If they are blank, and the incoming msg contains only msg.from or msg.to, then the node's default for the missing value is used just like it works now. You don't need to fill in both values in the incoming msg.

dceejay:
You're right of course. That is indeed how it works today.

I guess the question then is a) is that expected ? and b) if blank and someone omits to send in a msg.from or msg.to would suddenly getting a value that may be out of range they anticipated be useful/acceptable ?

juggledad:
Well since it is how it is currently working and no one has ever commented on it...I think it is a fine way to work (grin)

I can see a case where I want the from to be 1 all the time and I just want to send in the 'to' value. The thing that has cause me issues more than once is that I send in a msg.xxx to a node and it doesn't override the node's xxx property in the base nodes, but it does in other nodes (contrib nodes).

Since you have to edit the node to blank out (or set) one of the values, you should be aware of what you have done. ( Having the status show under the node showing the fron and to would give them a leg up on finding the issue.)

I'm changing the code after realizing that sending blank spaces in the msg.from or msg.to wasn't being handled properly but I think I have it resolved. Now if I can just get itrecoded, tested and put into the PR tonight before the grandkids come back tomorrow, I'll be happy. Once they arrive, Node-Red and anyone else's issues take a back seat (grin)

juggledad:
I've just posted the an updated version which checks for the blanks (actually makes sure the value is a number), added a status to show the range, and uses floor() and ceil() on the numbers when looking for an integer result

dceejay:
Hi - we don't want to set status for every single message passing through - If we have '000's of msgs per second then all that traffic has to go to the browser websocket. I think it should be set outside of the on input function - and just show the configured values or   just say "from msg.from to msg.to" etc if they are blank - which then also acts as a reminder to set them.

juggledad:
Hmm, when I put it before theh on input function
```
        this.status({fill:"red",shape:"dot",text:"random: from " + Number(this.low) + " to " + Number(this.high)})

        this.on("input", function(msg) {

```
it still  shows up each time a msg passes thru the node. I'm just going to remove it. Though I could put in a 'debug' option in the node to display the status if it was checked off....what do you think?

dceejay:
If outside the input then it will show up - but only once. So that should work.
And no - it is the debug - we aren't going to  add debug for debug options.

juggledad:
How about this? If the node 'from' and 'to' are blanked out, it will check the msg.from and msg.to to see if they are valid. If one or both is invalid, they will be reported in the error msg that will show in the debug log displaying the bad value
Screen Shot 2020-12-02 at 2 00 08 PM

PR #717, opened 2020-11-16
When multiple keys are entered at the same time, the contents of payload become abnormal.

t-kawamorita:
## Types of changes

## Proposed changes

Connect the "Pi Keyboard" node and the debug node and hit the two keys at the same time.
Then, the payload becomes an abnormal value as shown in the image.

The reason for the problem is that the PiKeyboardNode function does not support multi-line standard input.
I solved the problem by splitting the standard input, like the GPIOInNode function.

node-red-problem

## Checklist


dceejay:
Excellent - many thanks

t-kawamorita:
You're welcome.

PR #748, opened 2021-01-29
Join MUC when already connected

andreas-ibm:

## Proposed changes
Without this change, if we're already connected to the XMPP server and we try to join a new chatroom (MUC) then the node won't do anything (as the `presence` message is only sent in the `on('online')` callback.
Now we'll check if we're already connected and issue the `presence` message immediately to join the MUC.

## Checklist


andreas-ibm:
right, that commit list is just crazy, trying again

PR #757, opened 2021-02-18
timeswitch node: time zone support; do not mark as misconfigured when sunrise/sunset not used and lat/lon not given

pmacostapdi:
This pull request implements two changes to the timeswitch node:

a) support for specifying time zone when on time and/or off time are not specified as sunrise and/or sunset.

b) do not mark node as misconfigured if sunrise and sunset are not used and latitude/longitude are not given.


dceejay:
Thanks, looks interesting. Adding moment (that is now deprecated) - does blow this node up from 200k to 6.5MB - so you may want to see if there is a smaller (and supported) equivalent library.

pmacostapdi:
My bad, I replaced momentum with spacetime, which is maintained at much smaller.

time/timeswitch/timeswitch.js
```diff
@@ -14,6 +15,7 @@ module.exports = function(RED) {
         this.duskoff = n.duskoff;
         this.dawnoff = n.dawnoff;
         this.mytopic = n.mytopic;
+        this.timezone = n.timezone;
```
dceejay, line 18: Should this default correctly so that old flows work ? I thin line 17 will barf if it's undefined - should force default to UTC if not set

dceejay:
Yes - much better with spacetime... Thanks

time/timeswitch/timeswitch.js
pmacostapdi: Good catch! Yes, UTC is what the previous version of the node used, so I just checked in a version that makes that the default if undefined.

dceejay: great.

dceejay:
publish as v0.1.0 - many thanks for the PR.

PR #766, opened 2021-03-06
added i18n help files

heikokue:
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Proposed changes
separation of help-parts to separate en-US locales html files.
no content changes.

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red-nodes/blob/master/CONTRIBUTING.md)
- [x] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [ ] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #769, opened 2021-03-12
splitted datagenerator node help into locales

heikokue:
- [x] Preparatory change for further translations (no functional/feature change, no bugfix)

## Proposed changes
the headline says all

## Checklist


heikokue:
This is a splitted chunk of #766 and may replace it

dceejay:
Hi - yes happy to merge this - do you want to add the de translation file to it now or later ?

heikokue:
as you prefer! the de-json is ready. the de-html-help is still to do.
I created the split-PR as preparation for the further translation work, but which isn't finished yet.

what do you think about the need of a review in general?
the response and review support of the german community is quite low.
I would suggest to merge it even with no review.
Mistakes and further improvements can be also done after merge and driven by user feedback.
Even if I my translation work is 10% wrong, we'll have 90% improvement.
What do you think how to continue?

dceejay:
I agree re translation review - For these extra nodes they are much easier to patch / fix outside the core release cycles so any changes are much easier to make. So yes I'm happy to continue. I'm sure once we get some "wrong" translations out there then the German community will speak up :-)

PR #793, opened 2021-04-07
node-red-node-mysq add done() callback

hanc2006:
This fix adds the done() callback when the query has been executed.


dceejay:
Good spot - thanks

PR #796, opened 2021-04-12
added support for imap 'autotls' option to 'email in' node

HiroP0:
## Types of changes

If you want to raise a pull-request with a new feature, or a refactoring
of existing code, it **may well get rejected** if it hasn't been discussed on
the [forum](https://discourse.nodered.org) or
[slack team](https://nodered.org/slack) first.

HiroP0: Acknowledged. Yet I am not out for a discussion. Take it - even as a mere suggestion - or not. I hope this doesn't come across as too arrogant or ignorant, I just have little time and little experience with GitHub.

## Proposed changes

Some mail servers use a plain connection initially and later upgrade via STARTTLS command. This is fully supported by the node-imap package (https://github.com/mscdex/node-imap) underlying the node-red-node-email via the 'autotls' config option:

> **autotls** - string - Set to 'always' to always attempt connection upgrades via STARTTLS, 'required' only if upgrading is required, or 'never' to never attempt upgrading. Default: 'never'

The proposed change exposes the 'autotls' option to the 'email in' node.  

## Checklist

- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.  HiroP0: See above.
- [x] I have run `grunt` to verify the unit tests pass: HiroP0: after installing 500+ packes in about 15 operations: lint is fine, mocha finds 2/23 issues with email, none of which is related to what I did, AFAICS.
- [ ] I have added suitable unit tests to cover the new/changed functionality HiroP0: Sorry. I don't have a suitable-for-testing endpoint to test against and the coverage in email is patchy anyway (e.g. no POP)


HiroP0:
licence/cla signed meanwhile

hardillb (commented):
The comment about discussing it first is to help make sure that any changes in line with the projects plans also to help people making contributions don't end up doing things we can't use.

It doesn't need to be much, just a quick heads up so we know what's coming, it also helps with preventing merge clashes as we know if others are already doing work in the same area of the code.

The CLA needs to be signed with a github account that uses the same email address that made the commits against your git fork.

dceejay:
looks good to me - are you able to click/sign the CLA so we can merge ?

HiroP0:
In fact I did sign the CLA yesterday. If I follow the "Details" link it shows as being signed. The email matches with the commit. 

Yet the name does not. In the CLA I used my actual name, but not the GitHub account name. Any idea how to fix/reset this? (newbie there, as I said).

dceejay:
I think there is a way but I suspect it's easier to close this PR and just recreate with the correct ID :-)
Sorry it's a pain but

hardillb:
It should only be matching on the email address unless something has changed

dceejay:
yes matches email - but not if you use your actual name :-)

HiroP0:
By "closing PR" I guess you mean I submit another merge request with same content? (could well be this new request is then inheriting the CLA signing issue). But I will re-submit, if you think it's the way to go. Shall I?

dceejay:
Yes please

knolleary:
Closing and reopening PR won't do anything - it is the commits within the PR that potentially need redoing with the "right" email address.

When I look at the commit, it is recognised by GitHub - so whatever email you used to create the commit is associated with a github account. That is the account you need to use to sign the cla with.

dceejay:
thanks @knolleary - when are we moving to DCO ??? :-)

dceejay:
Thanks for this PR.

HiroP0:
Thanks for merging, which happened while I was writing this:
"Seems I cannot create a new pull request while the old one is open.
This is even the case after another change (I reverted the unnecessary line in JA l10n json). The exsiting pull request just gets enhanced."

dceejay:
:-) - the change was enough to kick the the system.. so hey.. all good.

PR #807, opened 2021-05-19
dynamic database name

man-of-fox:
Feature: database name can be dynamicaly provided as string oder context variable on "mongo in" and "mongo out"

## Proposed changes

missing feature to provide database name optionaly as context variable for example to use req.headers.host (without port) as database name. With that you can use same application with diferent data backends definded by query parameter, post data or in my case with domain name.

## Checklist


hardillb:
Do you have a link to where this change was discussed (Slack or Forum)?

PR #812, opened 2021-06-22
blink1: added "blink1 in" node

DennisGaida:
Has beend discussed in the Slack #core-dev channel.

    * sets msg.payload to current RGB value or "off" if RGB = 0,0,0
    * can also be used to pipe msg.payload to blink1 out
    * blink1 category created
    * blink1 label set to "blink1 out"
    * Bumped version to 0.1.0
    * Modified readme

## Types of changes

## Proposed changes

Currently only blink1 **out** is supported (i.e. set RGB colors). This change adds a blink1 **in** node (i.e. reads the current RGB colors)

## Checklist

Node wasn't included in any previous unit tests so I haven't added any.


hardware/blink1/package.json
```diff
@@ -13,7 +13,7 @@
     "keywords": [ "node-red", "blink1" ],
     "node-red"      : {
         "nodes"     : {
-            "blink1": "77-blink1.js"
+            "blink1": "blink1.js"
```
dceejay, line 16: can't change this - needs to be as-was.

hardware/blink1/blink1.js
```diff
@@ -4,7 +4,7 @@ module.exports = function(RED) {
     var Blink1 = require("node-blink1");
```
dceejay, line 4: Leave the file name as was or it will break.

hardware/blink1/blink1.html
```diff
@@ -1,5 +1,8 @@
-
 <script type="text/x-red" data-template-name="blink1">
+    <div class="form-row">
```
dceejay, line 2: Why move this ? usual default is last field as it's optional.

hardware/blink1/blink1.html
```diff
@@ -26,8 +25,9 @@
 
 <script type="text/javascript">
     RED.nodes.registerType('blink1',{
-        category: 'output',
```
dceejay, line 29: thought we said leave as output (for now at least)
(and input for the blink-in one)

hardware/blink1/blink1.js
```diff
@@ -4,7 +4,7 @@ module.exports = function(RED) {
...
     var blink1 = {};
 
-    function Blink1Node(n) {
+    function Blink1OutNode(n) {
```
dceejay, line 7: can't change name of existing node.

hardware/blink1/blink1.js
```diff
@@ -83,7 +83,52 @@ module.exports = function(RED) {
             node.error("No Blink1 found (" + e + ")");
         }
     }
-    RED.nodes.registerType("blink1",Blink1Node);
+    RED.nodes.registerType("blink1",Blink1OutNode);
```
dceejay, line 86: likewise can't change name of existing node

dceejay (changes requested):

hardware/blink1/blink1.html
DennisGaida: I checked multiple default nodes (Inject, Comment, Function) and mostly the name field is the first field in the list even though being optional.

hardware/blink1/blink1.html
DennisGaida: In chat you didn't say to leave it at output. I feel having "blink1 out" in category _output_ and "blink1 in" in _input_ makes it hard to find things. Right now I pretty much always have one category per "device" (e.g. Homeassistant = one category, Philips Hue = one category) and I feel blink1 should also be a category if more than one nodes exist. I don't have any other devices in "input" nor in "output".

But again this is opinionated. Go ahead and change it it doesn't hurt functionality.

hardware/blink1/package.json
DennisGaida: You said in Slack to change the file names if I want to. Now you are saying I shouldn't. What is it?

hardware/blink1/blink1.js
DennisGaida: It didn't break functionality for me? Existing nodes continued functioning even after changing the function name. Or is there another reason you say I can't change the name of existing node?

Again: Opinionated. I don't like the old name, but it doesn't matter functionality wise.

hardware/blink1/package.json
DennisGaida: I reread what you wrote - I misunderstood the filename comment. Will revert to 77-*

hardware/blink1/blink1.js
dceejay: if it isn't broken - don't change it. It is unnecessary for the new functionality.

dceejay:
Have merged.

but do you really mean to do this in the input ? `var payload = rgb.every(el => el === 0) ? "off" : rgb;`
ie send out a different result type if it is [0,0,0] ? Won't that make downstream processing non-trivial for everyone as you now have to check for a string first ?

DennisGaida:
> but do you really mean to do this in the input ? `var payload = rgb.every(el => el === 0) ? "off" : rgb;`
> ie send out a different result type if it is [0,0,0] ? Won't that make downstream processing non-trivial for everyone as you now have to check for a string first ?

I kind of stuck to what was initially there since you can send "off" to the _blink1 out_ node and it turns off (you could also send 0,0,0 and it would do the same). I only deviated from what was initially there with the returned array instead of the string since I thought array is nicer to work with downstream.

I documented the "off" vs array everywhere and right now I like it better this way since I can check for "off" more easily than checking an array containing only zeros in Node-Red.

dceejay:
hmm - yes ok - just wanted to make sure...  not sure about feeding from one to the other as indeed as you said forwarding [0,0,0] to it would do the same anyway... and surely it is (mostly) going to be reading back the state that the flow set it to so you ought to know how you set it already, and you shouldn't need to query at all (imho)  unless there is some 3rd party app that also sets it for you. Unless maybe the state in the device does survive a re-deploy so you could read it to initialise the flow ?

DennisGaida:
Dang thats some complicated thoughts there 👍 

I honestly didn't spend too much time thinking about it and just did what came to mind.

The "off" makes sense to me, since users of the Blink1 device might not know that 0,0,0 == off, personally I also don't like this and looking at a flow that checks "if msg.payload == 0,0,0 ..." in 5 years I wouldn't know why it checks for 0,0,0 and what that means.

But again: This is all a discussion and opinionated. You are the maintainer of the code so you have the last say. We can go all string (i.e. 0,0,0 or 255,0,0), we can go all array [0,0,0] or we keep the current mix. If we go for all string/all aray we might have to overthink the **out** node as well since you can write 0,0,0 AND "off" to it and it both works.

dceejay:
Well actually... if we look at the node-blink api there is actually a separate command for off (which we don't use) - so in theory I suppose 0,0,0 doesn't = off  - even though they say it does :-)...

Anyway - I haven't used this node for 4-5 years - and have never felt the need for an input node - so for now I'm OK to go with "off" - until we get the first complaint :-)...  and we can have the debate then. The other option is to make it a selectable option of the node - but meh.

DennisGaida:
I haven't used that thing in a long time either, but right now it comes in very handy as a busylight for people to know whether I am available or not.

PR #816, opened 2021-07-03
PiSrf: Added decimal places option + stop bugfix + more keywords

JsBergbau:

## Proposed changes

Added option to output more decimal places. When no decimal places are configured output stays the same for full backward compatibility.

Fixed bug that on restart flows error "Error stopping node: Close timed out" occured. Also removed sudo, because no need to run code with root rights. On systems where you need password for sudo it wouldn't run then. 

Added Keywords for "HC-SR04" and "SR04", because this module is fully compatible with this node.

## Checklist

### DIFF
--- a/hardware/PiSrf/nrsrf
+++ b/hardware/PiSrf/nrsrf
@@ -4,4 +4,4 @@ python_cmd='python3'
 command -v python3 > /dev/null || python_cmd=`python`
 
 BASEDIR=$(dirname $0)
-sudo $python_cmd -u $BASEDIR/nrsrf.py $@
+exec $python_cmd -u $BASEDIR/nrsrf.py $@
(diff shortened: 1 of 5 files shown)


dceejay:
functionality looks good - but is there any point is adding any more resolution than 1 dp ? Given that the datasheet for the SRF05 (which is supposed to be better than the 04) claims a resolution of 0.3 cm = 3.3 mm. Allowing any more just give a false impression of accuracy... (and indeed the actual accuracy seems to also varies with distance... so careful calibration would also be needed.  Some good testing done here - https://forum.arduino.cc/t/hc-sr04-tests-on-accuracy-precision-and-resolution-of-ultrasonic-measurement/236505

JsBergbau:
Good point. More than one dp resolution makes no sense. 
I'll change to either 0 or 1 decimal places.

JsBergbau:
Updated: Now only 0 or 1 decimal places are valid.

dceejay:
Many thanks for this update

PR #820, opened 2021-07-20
Update 66-mongodb.js

dodois:
- [ x ] New feature (non-breaking change which adds functionality)

## Proposed changes

Call done when the operation is complete

## Checklist

- [ x ] I have read the [contribution guidelines](https://github.com/node-red/node-red-nodes/blob/master/CONTRIBUTING.md)
- [ x ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [ x ] I have run `grunt` to verify the unit tests pass
- [  ] I have added suitable unit tests to cover the new/changed functionality


dceejay:
As the done capability was added later - you need to check if it exists before calling it in case someone is running a very old version that doesn't pass it it.

PR #823, opened 2021-07-26
Problems with more than one MCP3008

CD2468:
- [X] Bugfix (non-breaking change which fixes an issue)

## Proposed changes

It was only possible to communicate with one MCP3008 at once.
Now several MCPs can be used at the same time.  
+ Add the possibility to select CS 2 in the options.
+ The node displays now bus, device and cannel number 
+ Modified error message 

## Checklist

- [X] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [Link to Forum](https://discourse.nodered.org/t/problems-with-more-than-one-mcp3008/48860/11)

### DIFF
--- a/hardware/mcp3008/pimcp3008.js
+++ b/hardware/mcp3008/pimcp3008.js
@@ -1,12 +1,12 @@
 
 module.exports = function(RED) {
     "use strict";
-    var fs = require('fs');
+    var fs2 = require('fs');
     var allOK = false;
     var mcpadc;
     // unlikely if not on a Pi
     try {
-        var cpuinfo = fs.readFileSync("/proc/cpuinfo").toString();
+        var cpuinfo = fs2.readFileSync("/proc/cpuinfo").toString();
         if (cpuinfo.indexOf(": BCM") === -1) {
             RED.log.warn("Info : mcp3xxx : Not running on a Pi - Ignoring node");
         }
@@ -19,7 +19,7 @@ module.exports = function(RED) {
         RED.log.warn("Info : mcp3xxx : Not running on a Pi - Ignoring node");
     }
 
-    var mcp3xxx = [];
+    //var mcp3xxx = [];
 
     function PiMcpNode(n) {
         RED.nodes.createNode(this,n);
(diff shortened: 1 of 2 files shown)


hardware/mcp3008/pimcp3008.js
```diff
@@ -1,12 +1,12 @@
 
 module.exports = function(RED) {
     "use strict";
-    var fs = require('fs');
+    var fs2 = require('fs');
```
dceejay, line 4: Not sure why we need this duplication ?

hardware/mcp3008/pimcp3008.js
```diff
@@ -1,12 +1,12 @@
...
     var allOK = false;
     var mcpadc;
     // unlikely if not on a Pi
     try {
-        var cpuinfo = fs.readFileSync("/proc/cpuinfo").toString();
+        var cpuinfo = fs2.readFileSync("/proc/cpuinfo").toString();
```
dceejay, line 9: Likewise - why won't the original be ok ?

hardware/mcp3008/pimcp3008.js
```diff
@@ -28,23 +28,26 @@ module.exports = function(RED) {
         this.dnum = parseInt(n.dnum || 0);
         this.bus = parseInt(n.bus || 0);
         this.dev = n.dev || "3008";
+	this.fs = require('fs');
```
dceejay, line 31: indent ->

dceejay (changes requested):
Couple of tiny questions to fix but yes basically good...

hardware/mcp3008/pimcp3008.js
CD2468: This is a holdover from debugging.
The line can remain original

hardware/mcp3008/pimcp3008.js
CD2468: The line can remain also original.

hardware/mcp3008/pimcp3008.js
CD2468: Here I was not shure if the definition from the global place caused any problems. So I createt a new one and renamed the orginal to fs2.
If you want to remove the line you have also edit line 42.
I testet this on my raspberry and it worked
from:
`
this.fs.statSync("/dev/spidev"+node.bus+"."+node.dnum);
`
to:
`
fs.statSync("/dev/spidev"+node.bus+"."+node.dnum);
`
I testet this on my raspberry and it worked

PR #835, opened 2021-09-10
Node-red-node-twitter retweet support

### LINKED ISSUE #834: Node-red-node-twitter retweet support
Hi, the `node-red-node-twitter` is not supporting retweets, since it uses [`/statuses/update.json`](https://github.com/node-red/node-red-nodes/blob/96fa500ad663f68d49f456036421994d4ab25cf8/social/twitter/27-twitter.js#L640), and the retweet is done by [`statuses/retweet/:id`](https://developer.twitter.com/en/docs/twitter-api/v1/tweets/post-and-engage/api-reference/post-statuses-retweet-id).

I want to create a PR to add this support. I think since it is a different interface- it needs only the tweet id to retweet it, I will create a new node for it.

Or maybe the preferred way is to implement it inside the Twitter Out node?

WDYT?

I asked [in the forum](https://discourse.nodered.org/t/node-red-node-twitter-retweet-support/50038?u=baruchiro), but no answers there.

baruchiro:
Fixes #834

## Proposed changes

See #834. The `retweet` endpoint is different from the `update` endpoint, and I added it to the output node.

## Checklist

- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team:  
 **No answers [in the forum](https://discourse.nodered.org/t/node-red-node-twitter-retweet-support/50038?u=baruchiro), but we discussed on #834**.
- [ ] I have added suitable unit tests to cover the new/changed functionality:  
 **Unfortunately, there are no tests for Twitter Nodes**.

### DIFF
--- a/social/twitter/README.md
+++ b/social/twitter/README.md
@@ -59,3 +59,9 @@ To send a Direct Message (DM) - use a payload like.
 If `msg.media` exists and is a Buffer object, this node will treat it as an image and attach it to the tweet.
 
 If `msg.params` exists and is an object of name:value pairs, this node will treat it as parameters for the update request.
+
+#### Retweet
+
+To retweet an existing tweet, set `msg.retweet` to the tweet id. The tweet id can be found in the `id_str` field of the tweet object.
+
+Before retweeting, make sure the tweet is not a retweet itself.
--- a/social/twitter/27-twitter.js
+++ b/social/twitter/27-twitter.js
@@ -564,7 +564,7 @@ module.exports = function(RED) {
             });
 
             node.on("input", function(msg) {
-                if (msg.hasOwnProperty("payload")) {
+                if (msg.hasOwnProperty("payload") || msg.hasOwnProperty("retweet")) {
                     node.status({fill:"blue",shape:"dot",text:"twitter.status.tweeting"});
                     if (msg.payload.slice(0,2).toLowerCase() === "d ") {
                         var dm_user;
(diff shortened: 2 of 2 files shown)


social/twitter/README.md
```diff
@@ -59,3 +59,9 @@ To send a Direct Message (DM) - use a payload like.
 If `msg.media` exists and is a Buffer object, this node will treat it as an image and attach it to the tweet.
 
 If `msg.params` exists and is an object of name:value pairs, this node will treat it as parameters for the update request.
+
+#### Retweet
+
+To retweet an existing tweet, set `msg.retweet` to the tweet id. The tweet id can be found in the `id_str` field of the tweet object.
+
+Before retweeting, make sure the tweet is not a retweet itself.
```
baruchiro, line 67: For example, if you're using the Twitter Input Node with all the `#some_hashtag`, it will retweet the tweet with the `#some_hashtag`, and then, it will get it again, as a retweet.

If you will not filter it, it will try to retweet it again, and you will get an error "you already retweeted this".

dceejay: Can this check not be done within the node ?

baruchiro: How do you suggest doing that?

We can accept the whole `tweet` object. If we want to keep it simple with just `retweet` with the tweet ID, I guess we need to use the Twitter API to first check if the tweet ID has already been retweeted by us. I think it is complicated, and I wanted to start with simpler solution.

dceejay: If we don’t know how to do it then advising the user that they should is just passing the buck isn't it ?

knolleary: I would suggest changing the wording from:
```
Before retweeting, make sure the tweet is not a retweet itself.
 ```
to
```
You cannot retweet a retweet. You can check if it is a retweet using the `msg.tweet.XYZ` property.
```
(assuming these is a property under `msg.tweet` that does the job)

baruchiro: I will continue to work on it next week

baruchiro: > If we don’t know how to do it then advising the user that they should is just passing the buck isn't it ?

_Originally posted by @dceejay in https://github.com/node-red/node-red-nodes/pull/835#discussion_r705933923_

Not exactly. I don't know and I didn't research exactly what is the right way to validate we are not retweeting the same tweet twice, so I leave it to the user to implement a custom workaround, based on its use case.

For example, in my use-case, I'm retweeting tweets by hashtag, and I filtering all the tweets from me, so that it will never try to retweet twice, since the second tweet will be always by me.

baruchiro: > I would suggest changing the wording from:  
>```
>Before retweeting, make sure the tweet is not a retweet itself.
> ```
>to
>```
>You cannot retweet a retweet. You can check if it is a retweet using the `msg.tweet.XYZ` property.
>```
>(assuming these is a property under `msg.tweet` that does the job)

_Originally posted by @knolleary in https://github.com/node-red/node-red-nodes/pull/835#discussion_r705947899_

I'm very welcome to documentation improvements, my English is not so good. But about this specific suggestion, please see the previous comment. You **can** retweet a retweet, AFAIK, but you **can't** retweet tweet you already retweeted **yourself**. And I'm not sure how to check this (possibly by checking if one of the retweets of a tweets made by me...)

PR #851, opened 2021-11-02
Update package.jsons for mono-repo

sammachin:
Update the package.json for each node to point to the directory within the repo that hosts the code for the package in question.  as per https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository

This means that these node will now pass the repo-check in node-red-dev

I've not increased the version numbers as I'm not sure its worth publishing an update just for this change but these should find their way through to npm eventually with other updates.

### DIFF
--- a/analysis/mlsentiment/package.json
+++ b/analysis/mlsentiment/package.json
@@ -7,7 +7,8 @@
     },
     "repository" : {
         "type":"git",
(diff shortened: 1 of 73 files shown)


dceejay:
Most excellent - many thanks

PR #854, opened 2021-11-09
fix pool on acquire event cause MaxListenersExceededWarning

saknarak:
- [ X ] Bugfix (non-breaking change which fixes an issue)

### before
every query need to register pool on acquire event to specify `queryFormat` based on payload type
will cause `MaxListenersExceededWarning`

### after
from https://www.npmjs.com/package/mysql#pooling-connections

`pool.query` is a shortcut for `pool.getConnection()` -> `connection.query()` -> `connection.release()`

so use `pool.getConnection` and then
set `queryFormat` before `query` method be called

## Checklist

- [X] I have run `grunt` to verify the unit tests pass


dceejay:
Merged to test branch for using mysql2 library for wider testing... will release beta node later

dceejay:
@saknarak Published to npm as node-red-node-mysql@next for testing
will need manual install    `cd ~/.node-red && npm install node-red-node-mysql@next`
then restart

saknarak:
I just test with both `array` and `object` payloads with high volume messages.
Everything work as expect, no more  `MaxListenersExceededWarning` warning

dceejay:
Great, thanks for the PR and the feedback.
I would like some others to do some more testing on it first before I release it more widely as it is also moved to the mysql2 lib as that fixes the "old" auth problem with mysql, and seems more maintained, but does have a couple of minor differences. So there will be a short delay before it gets made into @latest.

Issue #811, opened 2021-06-21
node-red-node-mysql named parameters don't work anymore


lschloetterer:
I'm trying to use named parameters for a simple insert statement as described in the [example](https://flows.nodered.org/node/node-red-node-mysql#preparing-queries).
This has worked for me in the past.

Example here:
```json
[{"id":"c8088eb1.0da4a","type":"tab","label":"mysql-test","disabled":false,"info":""},{"id":"fe7d0da3.773d88","type":"inject","z":"c8088eb1.0da4a","name":"","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"{\"myText\":\"Hello\"}","payloadType":"json","x":130,"y":120,"wires":[["a2273968.e15a4"]]},{"id":"a2273968.e15a4","type":"template","z":"c8088eb1.0da4a","name":"prepared stmt","field":"topic","fieldType":"msg","format":"sql","syntax":"plain","template":"insert into \n  test.test \n  (`value`) \n  values \n  (:myText)","output":"str","x":320,"y":120,"wires":[["d37717cb.d93b78"]]},{"id":"d37717cb.d93b78","type":"mysql","z":"c8088eb1.0da4a","mydb":"e1c410b8.986518","name":"database","x":520,"y":120,"wires":[[]]},{"id":"41a140e5.6b7788","type":"inject","z":"c8088eb1.0da4a","name":"","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"{\"myText\":\"Hello\"}","payloadType":"json","x":120,"y":200,"wires":[["d6797222.7067b"]]},{"id":"d6797222.7067b","type":"template","z":"c8088eb1.0da4a","name":"mustache","field":"topic","fieldType":"msg","format":"sql","syntax":"mustache","template":"insert into \n  test.test \n  (`value`) \n  values \n  ('{{ payload.myText }}')","output":"str","x":290,"y":200,"wires":[["d37717cb.d93b78"]]},{"id":"e1c410b8.986518","type":"MySQLdatabase","name":"","host":"db","port":"3306","db":"test","tz":"Europe/Berlin","charset":"UTF8"}]
```

### Which node are you reporting an issue on?
node-red-node-mysql

### What are the steps to reproduce?
Use mysql node to try to insert rows with data from payload

### What happens?
[...]

PR #806, opened 2021-05-18
query by pool instead of using the connection directly

ericdum:
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Proposed changes

This PR is intended to increase the performance of Concurrent Request and to resolve some errors like CONNREST.
To do that by let MySQL Pool manage the connections instead of manually manage connections.

refer to: #804 

And FYI, there's some benchmark result below.
**Before optimize:** ab -n 500 -c 30
```
Document Path:          /data-flow/api/benchmark
Document Length:        13 bytes

Concurrency Level:      30
Time taken for tests:   13.528 seconds
Complete requests:      500
Failed requests:        0
Total transferred:      138500 bytes
HTML transferred:       6500 bytes
Requests per second:    36.96 [#/sec] (mean)
Time per request:       811.655 [ms] (mean)
Time per request:       27.055 [ms] (mean, across all concurrent requests)
Transfer rate:          10.00 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:       55  693 2318.4     88    9888
Processing:    59  102  34.5     93     204
Waiting:       59  102  34.5     93     203
Total:        117  795 2316.3    189    9995

Percentage of the requests served within a certain time (ms)
  50%    189
  66%    210
  75%    239
  80%    271
  90%    500
  95%   9902
  98%   9994
  99%   9994
 100%   9995 (longest request)
```

**After optimize:** ab -n 500 -c 30
```
Document Path:          /data-flow/api/benchmark
Document Length:        13 bytes

Concurrency Level:      30
Time taken for tests:   0.477 seconds
Complete requests:      500
Failed requests:        0
Total transferred:      138500 bytes
HTML transferred:       6500 bytes
Requests per second:    1048.70 [#/sec] (mean)
Time per request:       28.607 [ms] (mean)
Time per request:       0.954 [ms] (mean, across all concurrent requests)
Transfer rate:          283.68 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        8   12   1.6     12      17
Processing:     9   14   2.4     14      21
Waiting:        9   14   2.4     14      21
Total:         19   26   2.8     26      34

Percentage of the requests served within a certain time (ms)
  50%     26
  66%     27
  75%     28
  80%     29
  90%     30
  95%     32
  98%     32
  99%     33
 100%     34 (longest request)
```

## Checklist

[...]

PR #861, opened 2021-11-24
set query format function globally

ericdum:

## Proposed changes
if not do this,

when the `msg.payload` switching between array and object rapidly, 

it's will occurs sql parse error. 

## Checklist


ericdum:
> Hi - which issue is this about ? What is it trying to fix ? Why do we want to create a connection before anyone needs one ? Please can you target the 'test-mysql2` branch instead as that will be the next release.

if not do this,

when the msg.payload switching between array and object rapidly,

it's will occurs sql parse error.

ericdum:
> Hi - which issue is this about ? What is it trying to fix ? Why do we want to create a connection before anyone needs one ? Please can you target the 'test-mysql2` branch instead as that will be the next release.

and this pull request is derived from #806 , and #811 .

since 806 enhanced performance, and 811 resolved partial bug. 

this pull request is a patch after 811

dceejay:
moved to mysql2 branch

PR #863, opened 2021-11-29
Allow lat and lon to be 0

oliv3:

## Proposed changes

The node would raise a warning message if either lat or lon was 0.

## Checklist

Not familiar with javascript unit tests, sorry. Should be trivial to check if the node doesn't raise a warning when either lat or lon is 0, though :)


dceejay:
published as v0.1.10

PR #869, opened 2022-01-04
Sqlite Node - docs update

unborn-andy:
## Types of changes

Updated sqlite node README.md to match the more detailed node html file (the usage info was already there)
This can give better information to users that dont have the node installed but want to see how it works from the readme.md file.

Also added an example of how to use parameters Via msg.topic for both node html and README.md


dceejay:
Hi @unborn-andy - Any chance you can click/sign the CLA so we can accept and merge this PR ? Thanks

unborn-andy:
Hello @dceejay. I did the CLA.

By the way i just saw that @hardillb pushed a fix for the sqlite node parameters with the reseting of the `bind`. good job
(I didnt catch them on time to update the docs aswell) 

Hope this PR is ok .. its my first one ;)
Thanks

unborn-andy:
> did you want to update it before I merge it in light of the other PR ?

what do you mean ?
As you saw the changes i made dont affect the fix Ben did.
If we merge my PR does this mean we have to increment the node-red-node-sqlite version once more to 1.0.3 ?
(By the way i didnt change any version in package.json or anything in case its needed)

dceejay:
Hi - yes I will have to bump the version but that is OK... it's better that you don't bump the version and leave that for me - for exactly the reason you spotted - ie if I wanted to add multiple changes into the one release... 

I was just wondering if you wanted to add to the info.
(As it was just a fix then probably not - but I thought I saw something re % and $ that you wanted to clarify).

unborn-andy:
> I thought I saw something re % and $ that someone mentioned and you wanted to clarify

yes .. i asked in the[ Forum thread](https://discourse.nodered.org/t/sqlite3-data-insert-problem/56075/26) about this ..  
i was reading a bit more about the sqlite parameters and there are a few examples on the library's [Wiki page](https://github.com/mapbox/node-sqlite3/wiki/API#databaserunsql-param--callback). Ex.

```
      // Directly in the function arguments.
      db.run("UPDATE tbl SET name = ? WHERE id = ?", "bar", 2);

      // As an array.
      db.run("UPDATE tbl SET name = ? WHERE id = ?", [ "bar", 2 ]);

      // As an object with named parameters.
      db.run("UPDATE tbl SET name = $name WHERE id = $id", {
          $id: 2,
          $name: "bar"
      });
```

So based on the above examples .. we can pass params as named params with the $ syntax but ? is also valid.
In older versions of the node the ? systax was working (as reported by the OP) until you introduced that extra check in the code 
` if (msg.payload.length === (msg.topic.split('$').length - 1) ) { bind = msg.payload; }`

Maybe we can accept both ways of passing parms. If its worth it ;)
```
if ((msg.payload.length === (msg.topic.split('$').length - 1)) || (msg.payload.length === (msg.topic.split('?').length - 1)) ) {
    bind = msg.payload;
}

```

dceejay:
Ah no they are different - first passes an array of values the second passes and object with named properties.

PR #893, opened 2022-03-14
Special requirements for DietPi node-red installations

lscarneiro:

## Proposed changes

I wasted several hours investigating the `Error: getaddrinfo -3008` and couldn't solve it with any of the solutions proposed on this repo's issues nor on the dependency (`mdns` repo)[https://github.com/agnat/node_mdns].

In the end, it boiled down to DietP's minimalistic approach to default packages that don't even provide `avahi-daemon` out of the box (as you need to install it with `dietpi-software`) and looks like this wasn't enough for this node to work, I needed an extra prerequisite lib: `libnss-mdns`. 

I didn't try uninstalling `libavahi-compat-libdnssd-dev` or  `libudev-dev` to check if those are not needed for DietPi installations, but what solved in my case was installing this extra package.

I hope it helps someone in the future

## Checklist

### DIFF
--- a/io/mdns/README.md
+++ b/io/mdns/README.md
@@ -14,6 +14,15 @@ please read the [install instructions](https://www.npmjs.com/package/mdns) for t
 For Debian / Ubuntu this requires installing
 
         sudo apt-get install libavahi-compat-libdnssd-dev libudev-dev
+        
+> For DietPi installations, if you encounter this error when executing the node
+>
+>        Error: getaddrinfo -3008
+>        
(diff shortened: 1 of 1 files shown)


dceejay:
Hi @lscarneiro - was this just to support a particular node - or for the general installation ?
I don't think we need it in the general case - but as I don't use Diet-PI often I'm happy to be told otherwise.
If it is just a particular node then we can possible get that mentioned on it's install instructions.

Thanks

PR #898, opened 2022-04-04
Set text color on the RPi pinTable

bonanitech:

## Proposed changes

Set text color on the RPi pinTable so themes don't change it

## Checklist

### DIFF
--- a/hardware/PiGpio/36-rpi-gpio.html
+++ b/hardware/PiGpio/36-rpi-gpio.html
@@ -1,5 +1,6 @@
 <style>
     .rpi-gpio-pinTable {
+        color: #555;
         width: 340px;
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Hi @bonanitech 

Is there an issue this links to or at least a discussion on slack / forum for this issue?

Also, can you describe where you noticed this problem and why setting a constant color (vs using the node-red SASS variables) was chosen? (PS, you can paste screenshots into the thread to display the issue before and after)

bonanitech:
> If I used a SASS variable, the problem would persist because that color would change when generating the themes.

And I think creating a new variable just for that would be overkill

dceejay:
Thanks for that

PR #923, opened 2022-06-09
Smtp server

daviddhauwe:

## Proposed changes

Node-red-node-email more options for e-mail mta: make the smtp server more 'production ready'.
I've added the possibility to add TLS encryption or to use the Start TLS command. By default self-signed certificates are used, but it is possible to add your own certificates.
Authentication can be done with PLAIN or LOGIN and users can be defined.
Additional 'expert' options from <a href="https://nodemailer.com/extras/smtp-server/" target="_new">nodemailer SMTP server</a> can be made in as JSON.
The extra configurations have defaults so the node behaves exactly the same as before the change.
I did not add or change unit tests so it can be verified there are no breaking changes.

image

## Checklist

### DIFF
--- a/social/email/61-email.html
+++ b/social/email/61-email.html
@@ -61,45 +61,45 @@
 </script>
 
 <script type="text/javascript">
-(function() {
-    RED.nodes.registerType('e-mail',{
-        category: 'social-output',
-        color:"#c7e9c0",
-        defaults: {
-            server: {value:"smtp.gmail.com",required:true},
-            port: {value:"465",required:true},
-            secure: {value: true},
-            tls: {value: true},
-            name: {value:""},
-            dname: {value:""}
-        },
-        credentials: {
-            userid: {type:"text"},
-            password: {type: "password"},
-            global: { type:"boolean"}
-        },
-        inputs:1,
-        outputs:0,
-        icon: "envelope.png",
-        align: "right",
-        paletteLabel: function() {
-            return this._("email.email");
-        },
-        label: function() {
-            return this.dname||this.name||this._("email.email");
-        },
-        labelStyle: function() {
-            return (this.dname)?"node_label_italic":"";
-        },
-        oneditprepare: function() {
-            if (this.credentials.global) {
-                $('#node-tip').show();
-            } else {
-                $('#node-tip').hide();
+    (function () {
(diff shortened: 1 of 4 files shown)


dceejay:
Excellent - thanks

dceejay:
My only thoughts are....
1) when you select secure connection should the default port change to 465 ? (rather than remain 25 ?)
2) should all the extra options be hidden until you select secure connection ( as they may just confuse anyone who doesn't need them )

daviddhauwe:
@dceejay 
1. The default port is 1025 (I guess because on linux lower ports require root rights) - so changing this to 465 when selecting secure connection might be problematic too. In the e-mail node it is also not done. In the help, there is a note explaining ports and security. I propose to add a note like this: 
**In most cases when using port 465, check Secure connection. For port 587 or 25 keep it disabled, use Start TLS instead.**
2. I will do that: certificate and private key fields will be visible when secure connection or start tls is checked. When authenticate uses is checked, the users input will be visible.

dceejay:
1) true - yes please to note.
2) great - thanks - will be much neater.

coderofsalvation:
awesome work. But is there an easy way to change the json-textarea to a javascript-textarea?
I need to add an `inConnect` handler to block some spambots (but can't because the jsonvalidation fails)

dceejay:
No - we do not support JS as a typed input.

coderofsalvation:
ok thanks for getting back to me, I'll have to monkeypatch it another way then.

dceejay:
Hello ! and sorry - but opening up a general JS type input is just too fraught with possibilities for code injection.

coderofsalvation:
wait..are you now reasoning purely from a theoretical problem standpoint?
I just wanted to know if there's a JS type input supported in the ui-codebase for nodes.
To be clear: I don't really care about merging such patch upstream (I won't).
I just noticed there's a 'function'-node which allows JS but I noticed the textarea was declared in a totally different way than the json-textarea in this node.
If there's a quick workaround for this let me know.
Again, it is not my intention to steer the future-direction of this node, just to fix a local issue.

dceejay:
Well currently there is no input msg to that node... and once you are into hacking the code at that level you may as well just code in the callback directly :-)

Issue #944, opened 2022-08-22
Timeswitch reporting bizarre timestamp for next sunrise event


jdmallen:
### Which node are you reporting an issue on?
node-red-node-timeswitch

### What are the steps to reproduce?
1. Insert into flow
2. Set Lat & Long for your location
3. Configure to turn ON at SUNSET and OFF at SUNRISE
4. Deploy and observe scheduled time to next execution

### What happens?
The timeswitch node says it won't execute "until 24:10." The next event (sunrise, in my case) is actually more like 8.5 hours from now. Considering this is not a standard format for any timestamp, I suspect this is more of a presentation bug than an operational one, but I have no idea.

### What do you expect to happen?
I expect it to report that it'll send "off" either in the appropriate timespan for when sunrise will actually occur for my lat/long, or "until" an actual timestamp in my local time.

[BigTimer](https://flows.nodered.org/node/node-red-contrib-bigtimer), the node-contrib that this one references in [its overview page](https://flows.nodered.org/node/node-red-node-timeswitch), reflects the correct time. Please see these comparison pictures:

![image](https://user-images.githubusercontent.com/5761601/185825074-8a5b5cbe-5cef-4e03-9f7f-1b9f63661b3e.png)
![image](https://user-images.githubusercontent.com/5761601/185825154-106de08f-a6eb-4b11-8ec9-eb9ee7f3662e.png)
![image](https://user-images.githubusercontent.com/5761601/185825293-7e8ad80c-8a86-489c-8de5-6fc6636e15c2.png)
Note the "off until 24:10" under the timeswitch.

### Please tell us about your environment:

- [x] Node-RED version: v3.0.2
- [x] node.js version: Whatever it is in the latest [`nodered/node-red`](https://hub.docker.com/r/nodered/node-red/) Docker image.
- [x] npm version: "
- [x] Platform/OS: "
- [x] Browser: "

dceejay:
well yes definitely a presentation error at the very least... - 
would 24:10 (-> 00:10) be roughly the sunrise time in GMT rather than your local timezone ?

dceejay:
pushed version 0.2.1

jdmallen:
Wow, appreciate the quick work! I pulled 0.2.1 into NR and tried it again.

![image](https://user-images.githubusercontent.com/5761601/185909811-8fb4f056-6aab-4a1e-88fb-102f9e4359a7.png)


[...]

PR #945, opened 2022-08-23
Address incorrect timestamps bug in node-red-node-timeswitch; increase readability

jdmallen:

## Proposed changes

The purpose of this PR is primarily to address issue #944, wherein the timeswitch node was reflecting a time for sunset that simply didn't exist. It also was not showing the correct time for any event.

Applying this PR will:
- Fix the behavior of the node to address #944.
- Refactor the logic of the code to make better use of the [`spacetime`](https://github.com/spencermountain/spacetime) library that was already in the project.
- Remove most date arithmetic in lieu of `spacetime` functions.
- Simplify how next ON/OFF events are calculated.
- Track the next day's event time when today's has already passed. This makes sunrise/sunset events more accurate; e.g. it no longer uses today's sunrise time for tomorrow's sunrise, which is always different. It also solves the issue of midnight wrapping.
- Make use of lambda functions and newer ES features to increase readability/reusability of code.
- Add numerous comments to explain the logic and increase readability/maintainability.
- Apply minimal reformatting to a few lines.
- Add contributors to `package.json`.

Wishlist (not in this PR):

In future work on this library, I'd like to implement a few more improvements, after discussing them with the forum/Slack first:
- Move the lists of IANA timezones, selectable times, months, days of week, and any other constants to a separate JSON file for storage and read on load.
- Eliminate the switch statements for something a bit simpler. Maybe an array of ints representing months or days of week, provided by the JSON, that can be filtered by what's selected in the UI.
- Move more of the inline <script> stuff to the JS file.

## Checklist

- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team. _(This is a bugfix.)_
- [ ] I have added suitable unit tests to cover the new/changed functionality. _(No new/changed functionality.)_

### DIFF
--- a/time/timeswitch/package.json
+++ b/time/timeswitch/package.json
@@ -25,7 +25,19 @@
     },
     "contributors": [
         {
-          "name": "@pmacostapdi"
+            "name": "@dceejay"
+        },
+        {
+            "name": "@pmacostapdi"
+        },
+        {
+            "name": "@heikokue"
+        },
+        {
+            "name": "@sammachin"
+        },
+        {
+            "name": "@jdmallen"
         }
     ]
 }
--- a/time/timeswitch/timeswitch.js
+++ b/time/timeswitch/timeswitch.js
@@ -3,6 +3,8 @@ module.exports = function(RED) {
     "use strict";
     var SunCalc = require('suncalc');
     const spacetime = require("spacetime")
+		const SUNRISE_KEY = "sunrise";
+		const SUNSET_KEY = "sunset";
 
     function TimeswitchNode(n) {
         RED.nodes.createNode(this, n);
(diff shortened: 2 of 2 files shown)


jdmallen:
Note: not all my commits individually result in a functioning node. If this PR is accepted, I recommend performing a squash+rebase onto master, rather than retaining the individual commits, if that's not the standard.

dceejay:
Works fine for me also. So, happy to merge as-is. 
Up to you if you want to work on the extras you mention above. I wouldn't probably worry about moving the timezone list out though.

dceejay:
Also - is it worth moving to spacetime 7.1.4 ?

jdmallen:
Thank you! Funny enough, I actually had the new version of spacetime in there before I reverted the file to minimize to only necessary changes. I'm not sure if there's anything we use in spacetime that merits the bump to the next major version, but I'll certainly look at it for future work.

dceejay:
Yes always useful to try to keep it current while other things are being fixed as long as it's not a major re-write... just in case of vulnerabilities etc.  but no imperative at the moment.

Thanks again. (no pushed as v 0.2.2)

PR #954, opened 2022-10-07
Twitter API V2.0 public search update

shahramdj:

## Proposed changes

Current twitter node does not search public tweets since twitter has upgraded the API to V 2.0. The change that I made solves the problem.
## Checklist

- [X] For non-bugfix PRs, I have discussed this change on the forum/slack team.


hardillb:
You should not check in the `node_modules` directory, please update your local branch to remove it.

social/twitter/27-twitter.js
```diff
@@ -11,77 +11,126 @@ module.exports = function(RED) {
...
-                this.localIdentityPromise = Promise.resolve(localUserCache[self.credHash]);
-            } else {
-                this.localIdentityPromise = this.get("https://api.twitter.com/1.1/account/settings.json").then(function(body) {
-                    if (body.status === 200) {
-                        localUserCache[self.credHash] = body.body.screen_name;
-                        self.screen_name = body.body.screen_name;
-                    } else {
-                        self.warn("Failed to get user profile");
+
+            const needle = require('needle');
```
Steve-Mcl, line 51: I see here you are using `needle` however `needle` is not a dependency in this package (`request` is a dependency).

I am not against using `needle` (in fact it is a reasonable alternative considering `request` is depreciated) BUT for me to accept this PR, I would suggest the rest of the HTTP requests are changed to use `needle` as well. 

If you can do that - great. 

PS, you would also need to add a dependency in the package.json for `needle`

social/twitter/27-twitter.js
```diff
@@ -198,111 +246,280 @@ module.exports = function(RED) {
...
+                                    await deleteAllRules(currentRules);
+                                    // Add rules to the stream. Comment the line below if you don't want to add new rules.
+                                    await setRules();
+                                } catch (e) {
+                                    console.error(e);
+                                    process.exit(1);
+                                }
+                                // Listen to the stream.
+                                // node.status({fill:"green", shape:"dot", text:(node.tags||" ")});
+                                console.log("Twitter API is steraming public tweets with search term " + node.tags || " ");
```
Stephen-Broadhurst, line 424: Love the project to upgrade to Twitter 2.0.  Would be great to keep the logging consistent with the old node, and avoid console.log() throughout, so node.debug maybe here?

social/twitter/27-twitter.js
```diff
@@ -631,31 +848,31 @@ module.exports = function(RED) {
...
                         }
-                        mediaPromise.then(function(mediaId) {
+                        mediaPromise.then(function (mediaId) {
                             var params = msg.params || {};
                             params.status = msg.payload;
                             if (mediaId) {
                                 params.media_ids = mediaId;
                             }
-                            node.twitterConfig.post("https://api.twitter.com/1.1/statuses/update.json",{},params).then(function(result) {
+                            node.twitterConfig.post("https://api.twitter.com/1.1/statuses/update.json", {}, params).then(function (result) {
```
Stephen-Broadhurst, line 857: So there is this mix of twitter 1.1 for the Twitter Out node, and 2 for the Twitter in node.  It would be fantastic to have them both on Twitter 2.  At the moment if I use this branch and deploy first a twitter in, and then a twitter out node I get an exception I think because the two configurations of credentials clash `24 Oct 22:45:40 - [info] Updated flows
24 Oct 22:45:40 - [red] Uncaught Exception:
24 Oct 22:45:40 - [error] Error: aborted
    at connResetException (node:internal/errors:705:14)
    at TLSSocket.socketCloseListener (node:_http_client:454:19)
    at TLSSocket.emit (node:events:525:35)
    at node:net:301:12
    at TCP.done (node:_tls_wrap:588:7)`

social/twitter/27-twitter.js
shahramdj: Good suggestion, I will change it.

hardillb:
Might want to wait until after the 9th to see what API is still available and at what cost...

BeeOnLion:
Hi all

I was wondering if there is any update as to weather or not this is going to be merged into master or is there a timeline on this? 

@shahramdj did you manage to get this running with the current version 2 of the twitter API? Or have you reverted back to using a function node to get your tweets out to the world?

PR #968, opened 2023-01-11
I need error msg

tamasvar:
I still need the msg.payload when there is an error, not only when everything is going well.

## Checklist

### DIFF
--- a/storage/mysql/68-mysql.js
+++ b/storage/mysql/68-mysql.js
@@ -142,6 +142,8 @@ module.exports = function(RED) {
                             conn.query(msg.topic, bind, function (err, rows) {
                                 conn.release()
                                 if (err) {
+                                    msg.payload = rows;
+                                    send(msg);
                                     status = { fill: "red", shape: "ring", text: RED._("mysql.status.error") + ": " + err.code };
                                     node.status(status);
                                     node.error(err, msg);


dceejay:
I can understand why you may want it - but for naive users I think it is better to not send anything when there is an error, rather than send a (normally) empty [] response that would look like a valid response unless you check for an error. So I'm not (yet) convinced that this a change we should make. You could also use the error to create a dummy payload [] if you need to forward it on - but yes your fix would do that for you.

shaqaruden:
> I can understand why you may want it - but for naive users I think it is better to not send anything when there is an error, rather than send a (normally) empty [] response that would look like a valid response unless you check for an error. So I'm not (yet) convinced that this a change we should make. You could also use the error to create a dummy payload [] if you need to forward it on - but yes your fix would do that for you.

I agree that it shouldn't send an error but I do think it should throw an error

jonbdk:
I would suggest a slightly different approach - with largely the same result. 
Create a secondary output, which is used in case of error.
In 64-mysql.html change:
`outputs:1,`
to
`outputs:2,`
In 68-mysql.js
After the line (147)
`node.error(err, msg);`
add line:
`send([null, { query: msg.topic, payload: 'ERROR', error: err }]);`

You could event remove the node.error() - if you wanted it to be less chatty

phire-tc:
I consider myself a naive user yet I can't imagine why I would want my flows to just stop when there's a problem. Is there any way that my code can be notified of the error and do something in response?

dceejay:
It does send an error - it can be caught with the catch node - which is the standard way we do it for most nodes.

PR #997, opened 2023-05-16
Bugfix all sense channels

maxfoxim:
Hello,

this pull request fixes a small bug that avoids reading envrioment and motion group at the same time. Without the bugfix it is only possible to read one.

The bug is discussed here:
https://discourse.nodered.org/t/not-possible-to-read-all-values-in-sensehat-node/78052/13

- [] New feature (non-breaking change which adds functionality)

## Proposed changes

One has to set **lf_enabled = False** to **lf_enabled = True**. This enables the reading of both in the global process command.

### DIFF
--- a/hardware/sensehat/sensehat.py
+++ b/hardware/sensehat/sensehat.py
@@ -66,7 +66,7 @@ hf_interval = 0.09 # Approx 10/s
 lf_interval = 1
 
 hf_enabled = False
-lf_enabled = False
+lf_enabled = True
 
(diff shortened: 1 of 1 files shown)


maxfoxim:
Hey @dceejay 

I just signed it. Hope it works now.

Best, Stephan

dceejay:
Seems like you signed the commit with a different ID than the one you used for the CLA. they need to be the same...

Issue #330, opened 2017-06-07
SerialPort: programmatically close/reopen


dvv:
Hello!

I have SerialPort <-> MQTT as part of my flow and I would like to have means for the portIn and portOut nodes to temporarily release its ttyUSBx so that avrdude could update the firmware of connected device.

So far I do manually assign different (dummy) port to portIn/Out nodes, apply changes, manualy flash then restore node assignments.

Any pointers?

TIA

dceejay:
We have no plans to do this from within a flow.
We could consider a Pull request if anyone wanted to build one.

PR #1012, opened 2023-06-18
Serial port connect/disconnect, configurable reconnect time

htmltiger:

## Proposed changes

Auto reconnect time changed from hardcoded to configurable.

Add option to disconnect/reconnect serial port on demand.
https://github.com/node-red/node-red-nodes/issues/330

I have updated locale for english, other langs will need to be updated.

## Checklist


Steve-Mcl:
Hi @htmltiger thank you for taking the time to contribute. It is appreciated.

We normally prefer discussion before making a PR so that it aligns with the goals of the project and avoids unnecessary rework or rejection. 

For example, new features should be based off the dev branch and should therefore target the Dev branch. So the first thing you will need to do is to rebase this PR.

Second, #330 was from 2017 - a lot has changed since then. For example, the MQTT nodes now have the  runtime control capability (connect/disconnect) and the ability to set the connection properties too. This is done via a `msg` property named `action` where action is a string of "connect", "disconnect" etc. This would have been my preferred method of control since it aligns with how MQTT does it and permits a future modification to include serial setup parameters in another property along side the action property like ".connection" or ".configuration".

htmltiger:
I have changed connect/disconnect to action to align with mqtt. Thanks for the hint.
I could not find dev branch to rebase it.

I will leave it here for anyone who finds it helpful as i have searched in issues/pr for the same features and f for my use and had to modify it so it was necessory for me and possibly for many others who have arduino connected to raspberry pi and want to flash it over ota using avrdude without stopping and starting node-red.

Just in case if this does not get merged to official and anyone stumbling here wants to install this:
manually copy `serialport` from my repo as `~/.node-red/serialport` to locally and run 
`cd ~/.node-red` 
uninstall original `npm remove node-red-node-serialport`
`npm install ~/.node-red/serialport` 
so locale will be displayed or it will be a blank config text if copied to .node-red/nodes

io/serialport/25-serial.html
```diff
@@ -245,7 +250,8 @@
             bin: {value:"false"},
             out: {value:"char"},
             addchar: {value:""},
-            responsetimeout: {value: 10000}
+            responsetimeout: {value: 10000},
+            serialReconnectTime: {value: 15000}
```
dceejay, line 254: Can't set this here - as then settings.js would never be able to override... must default to blank

dceejay:
Thanks for this - as Steve said we usual prefer some discussion first - as indeed there are two separate things here you are looking at fixing so two PR would have been easier to discuss.

So - re the point Steve made - I agree a single action property is cleaner as otherwise you could have specified both msg.connect and msg.disconnect on the same message which could be confusing...

Also (I'm on vacation so haven't had a chance to dig into the code yet) - what happens if you send a msg to be sent - once you have disconnected ? hopefully it is handled gracefully ? 

If you then do disconnect - does the reconnect timer then kick in and try to reconnect ? Is there a flag to stop it retrying ?

Then re the 2nd part of the PR - re reconnection. You can't set the default in the node to be 15s - as then the settings.js would not be able to set the default. Any new nodes should be set with a blank (falsey) value so that the settings.js value has priority.

io/serialport/25-serial.js
```diff
@@ -483,6 +500,24 @@ module.exports = function(RED) {
                 else {
                     done();
                 }
+            },
+            connect: function(port) {
+                if (connections[port]) {
+                    try {
+                        connections[port].connect();
+                    }
+                    catch(err) { }
```
dceejay, line 509: Do we really want to do nothing ? should this raise an error - or set node.status at least ?

io/serialport/25-serial.js
```diff
@@ -483,6 +500,24 @@ module.exports = function(RED) {
...
+                }
+            },
+            disconnect: function(port) {
+                if (connections[port]) {
+                    try {
+                        connections[port].disconnect(function() {
+                            RED.log.info(RED._("serial.errors.closed",{port:port}), {});
+                        });
+                    }
+                    catch(err) { }
```
dceejay, line 519: Do we really want to do nothing ? should this raise an error - or set node.status at least ?

io/serialport/25-serial.html
htmltiger: I have updated it, now its an empty string.

io/serialport/25-serial.js
htmltiger: try catch was not necessory as we are using existing functions to connect so it should be handled in there.

io/serialport/25-serial.js
htmltiger: same here, try catch was not necessory as we are using existing function to disconnect so it should be handled in there.

htmltiger:
> Thanks for this - as Steve said we usual prefer some discussion first - as indeed there are two separate things here you are looking at fixing so two PR would have been easier to discuss.

Sure, will keep this in mind for future.

> 
> So - re the point Steve made - I agree a single action property is cleaner as otherwise you could have specified both msg.connect and msg.disconnect on the same message which could be confusing...

done.

> 
> Also (I'm on vacation so haven't had a chance to dig into the code yet) - what happens if you send a msg to be sent - once you have disconnected ? hopefully it is handled gracefully ?

I have checked it and now and there were no errors on serial out node, not sure about request node.

> 
> If you then do disconnect - does the reconnect timer then kick in and try to reconnect ? Is there a flag to stop it retrying ?

I have added auto connect checkbox (default true as before) and on manual disconnect it stops auto reconnect, on connect it restores old checkbox value.

> 
> Then re the 2nd part of the PR - re reconnection. You can't set the default in the node to be 15s - as then the settings.js would not be able to set the default. Any new nodes should be set with a blank (falsey) value so that the settings.js value has priority.

I have changed it as you have requested but the input field will be blank for now, 
We need a way to get settings.serialReconnectTime as a placeholder so its easy to understand the default if nothing is entered.

Steve-Mcl:
@htmltiger sorry I lost track of this. Do you feel it is ready to go?

@dceejay any thoughts moving this forward please?

htmltiger:
@Steve-Mcl  yes, it has been running fine since the last changes and its ready to merge.

Steve-Mcl:
Thanks for the feedback. 

I will review soon (tomorrow / monday) but ultimately, I would want Daves sign off before merging.

dceejay:
Yes, looks better now. My comments have been addressed. (Now just uses action: connect   action: disconnect)  
So yes - I think we can go ahead.
We should also look to see what underlying serial port npm (now v11) brings and possible update that before release.

htmltiger:
@dceejay I have tried to import `RED.settings.serialReconnectTime` in `25-serial.html` for placeholder but its not getting the settings.js value as its not available to the browser
[var defReconnectTime = RED.settings.serialReconnectTime || 15000;](https://github.com/htmltiger/node-red-nodes/blob/d73bb47ba641089ade8f0cfe8efb13b2824358df/io/serialport/25-serial.html#L268)

Steve-Mcl:
Hi @htmltiger great work, but, a few small issues preventing me accepting this PR.

1. On my testing, the autoconnect checkbox seems to be ignored:

https://github.com/node-red/node-red-nodes/assets/44235289/3b8ddb24-accc-47cf-9be7-edf223f07bb8


2. When disconnected, messages are queued and never released. @dceejay this looks a like a bug waiting to happen.  There is nothing in the code to dequeue on connect/re-connect. 
    1. So, there is a choice to make here:
        1. Drop messages (dont queue) while disconnected
        1. Dequeue messages upon connect/reconnect

https://github.com/node-red/node-red-nodes/assets/44235289/0fd25a4f-21d7-4a91-8993-5d4d69635fd2


3. We dont pre-populate the timeout 
    * I realise you have mentioned this and I dont think it was explained how to get that backend value to client side, but I think the placeholder should be set by this value (not the actual value, only the placeholder)
    * The setting can be made avaiable client side by adding `safeSettings.serialReconnectTime = runtime.settings.serialReconnectTime`  in the file `packages/node_modules/@node-red/runtime/lib/api/settings.js` around line `169`
    * image
    * image



---

### Visual stuff

A tiny nitpick (_a me thing_) 

The layout of the autoconnect check and the reconnect time are a bit too bunched up for my OCD & the row has no icon.  

#### Current PR
image


#### Proposed
image




PS: My apologies for not getting this reviewed properly sooner.  

PS2: If you are not in a position to progress the above please let me know and I will endeavour to take over and get them added to your great work.

dceejay:
Hi @htmltiger - I totally agree with @Steve-Mcl about the layout - Visual Stuff above

and yes re 1 - to me it seems to not honour the flag on deploy - so if you redeploy and it is set to not auto connect - it does so anyway. 

re 2 - my thinking is that we should drop any messages - as we would also be dropping any incoming ones as we are not connected at that point. Managing queues for messages and retries is a whole different area (and different set of nodes :-)

htmltiger:
@Steve-Mcl feel free to take over or submit changes in this pr.
Thanks.

dceejay:
Solved this another way by adding a serial control node that can stop, re-configure, and re-start a serial port. See PR #1035

PR #1018, opened 2023-07-20
Support stop msg for daemon node

gmarzot:
## Types of changes

## Proposed changes

This PR adds support to the node-red-node-daemon for a stop message. When stopped the process will be terminated as per the configured signal and will not be restarted by redo until re-set via a subsequent start message. 

Documentation updated on start message to indicate it will not restart a running process

### DIFF
--- a/utility/daemon/daemon.js
+++ b/utility/daemon/daemon.js
@@ -34,7 +34,7 @@ module.exports = function(RED) {
         function inputlistener(msg) {
             if (msg != null) {
                 if (msg.hasOwnProperty("stop")) {
-                    this.stopped = true;
+                    node.stopped = true;
                     if (node.running) {
                         node.child.kill(node.closer);
                     }
@@ -46,7 +46,7 @@ module.exports = function(RED) {
                     node.child.kill(msg.kill.toUpperCase());
                 }
                 else if (msg.hasOwnProperty("start")) {
-                    this.stopped = false;
+                    node.stopped = false;
                     if (!node.running) {
                         let args = "";
                         if (msg.hasOwnProperty("args") && msg.args.length > 0) {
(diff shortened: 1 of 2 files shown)


utility/daemon/daemon.js
```diff
@@ -32,16 +33,30 @@ module.exports = function(RED) {
...
-                        args = parseArgs(msg.args.trim());
+                else if (msg.hasOwnProperty("start")) {
+                    node.stopped = false;
+                    if (!node.running) {
+                        let args = "";
+                        if (msg.hasOwnProperty("args") && msg.args.length > 0) {
+                            args = parseArgs(msg.args.trim());
+                        }
+                        runit(args);
+                    } else {
```
dceejay, line 56: Why are we doing this here ? If it is called with start and is already running then we don't want to kill it do we ? Surely just leave it running.

utility/daemon/daemon.js
```diff
@@ -32,16 +33,30 @@ module.exports = function(RED) {
 
         function inputlistener(msg) {
             if (msg != null) {
-                if (msg.hasOwnProperty("kill") && node.running) {
+                if (msg.hasOwnProperty("stop")) {
+                    node.stopped = true;
+                    if (node.running) {
+                        node.child.kill(node.closer);
+                    }
+                    node.debug(node.cmd+" stopped by msg");
```
dceejay, line 41: Don't think we need extra debug.

gmarzot:
i added the kill on the start msg to trigger a restart, to match documented
behavior..

but equally valid (or more so i guess) would be to change the doc to say
that msg.start only starts (no restart behavior). I can make that change.

eh... debug to cover each message type.. i see no harm as it is off
normally.. your call.

-G

gmarzot:
one lingering thought is that if someone wants to add new args and trigger
a  restart.. the current PR supports that w/ a single msg...

but that can be accomplished with stop then start w/ args.. so i am
continuing to remove the kill i added to msg.start

dceejay:
Thanks for the PR.

PR #1021, opened 2023-08-07
STOMP: Improved error stated and logging

OlivierVerhaegen:
## Types of changes

## Proposed changes

Adds better reflection of the connection status using the status of the STOMP nodes. Currently all nodes will have an error state which doesn't reset after any error occurs. With this update this will only happen when the client is disconnected from the STOM server. If another error occurs only a message is logged but the connection state is not changed as the client is still connected to the server.

An example: "unexpected ACK message from client".
This error occurs when you send an ACK message to the server without setting up the correct acknowledgement mode on subscription. 
-> no need to set an error state across all STOMP nodes, logging this error (also in debug window) is sufficient as this is no connection issue which leads to other nodes failing. 

## Checklist

### DIFF
--- a/io/stomp/18-stomp.js
+++ b/io/stomp/18-stomp.js
@@ -219,11 +219,10 @@ module.exports = function(RED) {
                     });
 
                     node.client.on("reconnecting", function() {
-                        node.warn("Reconnecting to STOMP server");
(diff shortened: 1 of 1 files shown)


dceejay:
Excellent - thanks again

PR #1035, opened 2023-11-14
node-red-node-serialport port selection node addition

yhur:

## Proposed changes

When the node-red starts, the flow(program) picks up the hard-coded serial ports and open them. But when a device re-connects after disconnecting for any reason, it may be possible the port number change, but the end user of the flow can't change the port unless he/she has the access to the node-red editor.

Recently I had to develop a feature on top of node-red-node-serialport that enables the end user of node-red to change the port number/name without the access to the node-red editor.

This PR adds the said functionality to the existing node-red-node-capability

********** This version: IMAP ONLY **********

## Checklist


yhur:
@dceejay,
As discussed, here is the PR for the modification. Thanks.

dceejay:
Hi - finally had a chance to check this out... looks like a good idea - but... if we are going to do this then I think it should offer more control over the port. As we have another PR outstanding about stop start control and this would be an opportunity to do both in a not-intrusive manner. 
IE can we also allow changing baud rate, (and maybe data bits and parity ) ?
Also offer stop (flush/close) , and start control also.

Also ned some help docs for the info tab (happy to help with that).

io/serialport/25-serial.js
```diff
@@ -27,99 +27,115 @@ module.exports = function(RED) {
         this.out = n.out || "char";
         this.waitfor = n.waitfor || "";
         this.responsetimeout = n.responsetimeout || 10000;
-    }
-    RED.nodes.registerType("serial-port",SerialPortNode);
+    
+        this.newPort = (serialPort) => {
+            serialPool.close(this.serialport,() => {});
```
dceejay, line 32: If the port has failed to open (if it's not there for example) - then will close work OK ? And if it is open then should we flush any unwritten data first before close ?

io/serialport/README.md
```diff
@@ -55,3 +55,22 @@ This node behaves as a tightly coupled combination of serial in and serial out n
...
+    "port": "/dev/tty.usbmodem1234561",
+    "serialbaud": 115200,
+    "databits": 8,
+    "parity": "none",
+    "stopbits": 1
+} 
+```
+
+
+https://github.com/yhur/node-red-nodes/assets/13171662/5d5f2e7b-0cc8-421c-ad05-6361a4ececdd
```
dceejay, line 75: Sorry - no links to external resources please.

io/serialport/README.md
```diff
@@ -55,3 +55,22 @@ This node behaves as a tightly coupled combination of serial in and serial out n
 Send the request message in `msg.payload` as you would do with a serial out node. The message will be forwarded to the serial port following a strict FIFO (First In, First Out) queue, waiting for a single response before transmitting the next request. Once a response is received (with the same logic of a serial in node), or after a timeout occurs, a message is produced on the output, with msg.payload containing the received response (or missing in case if timeout), msg.status containing relevant info, and all other fields preserved.
 
 For consistency with the serial in node, msg.port is also set to the name of the port selected.
+
+### Port Select
```
dceejay, line 59: This info (plus any changes ;-) needs to be copied to the sidebar info tab as well. locales/en-US/25-serial.json

io/serialport/25-serial.js
```diff
@@ -200,6 +224,32 @@ module.exports = function(RED) {
...
+        this.serial = n.serial;
+        var node = this;
+        node.port = serialPool.get(this.serialConfig);
+        node.on("input",function(msg) {
+            if (msg.payload.hasOwnProperty("port")) { 
+                node.serialConfig.newPort(msg.payload);
+                node.serialConfig.emit('newport',msg.payload);
+                node.send(msg);
+            } else {
+                node.send({payload: {port: node.serialConfig.serialport}});
```
dceejay, line 247: should it send complete current port config ? baudrate , etc ?

io/serialport/25-serial.js
```diff
@@ -200,6 +224,32 @@ module.exports = function(RED) {
...
+        }
+
+        this.serial = n.serial;
+        var node = this;
+        node.port = serialPool.get(this.serialConfig);
+        node.on("input",function(msg) {
+            if (msg.payload.hasOwnProperty("port")) { 
+                node.serialConfig.newPort(msg.payload);
+                node.serialConfig.emit('newport',msg.payload);
+                node.send(msg);
```
dceejay, line 245: as per below - should this not be the complete new config ?

yhur:
@dceejay,
Thank you for your feedback. 

I modified according to your feedback and implemented the stop/start as well and it is now renamed to 'serial control'

io/serialport/25-serial.js
```diff
@@ -200,6 +224,42 @@ module.exports = function(RED) {
...
+        if (!this.serialConfig) {
+            this.error(RED._("serial.errors.missing-conf"), {});
+            return;
+        }
+
+        this.serial = n.serial;
+        var node = this;
+        node.port = serialPool.get(this.serialConfig);
+        node.on("input",function(msg) {
+            if (msg.payload.hasOwnProperty("config")) {
```
dceejay, line 242: I think this is now too complicated...  I think whatever the input is, the node should just reply with the (new) config - I.e. the current config if nothing changes - or the updated one if something has changed. No need for a config property - just the payload as you had it before.... (but see below)

io/serialport/25-serial.js
```diff
@@ -200,6 +224,42 @@ module.exports = function(RED) {
...
+        node.on("input",function(msg) {
+            if (msg.payload.hasOwnProperty("config")) { 
+                if (msg.payload.config === "query") { 
+                    node.send({payload: node.serialConfig});
+                } else {
+                    node.serialConfig.changePort(msg.payload.config);
+                    node.serialConfig.emit('start',msg.payload.config);
+                    node.send({payload: msg.payload.config});
+                }
+            } else if (msg.payload.hasOwnProperty("start")) {
```
dceejay, line 250: Rather than two separate properties - I'd prefer a single "enable" (or control ?)  property that is set to true or false and maybe also allows start and stop and enable and disable ?

dceejay:
Hmm - not sure I like a separate config property - I prefer the way you had that before.. just add the properties to the msg.payload level.
And just to add a single "enable" or "control" property that can be true/false etc ?
But much closer.

yhur:
I originally started with a very simple requirement of changing the serial port, and taking your stop/start feedback into the code(btw, I like your requirement), it is now more full blown feature, I guess.

When adding the stop/start, I changed my view on the message from the configuration property to the command, because stop/start sounds like a command, so the initially thought property had to be a command, so it turned into the 'config' command. Though the stop and start command looked somewhat awkward with no values.

But since it is suggested to see the message as the properties, it now looks to have the better shape and consistency.

io/serialport/25-serial.js
```diff
@@ -200,6 +222,45 @@ module.exports = function(RED) {
...
+                node.serialConfig.enable = msg.payload.enable;
+                if (msg.payload.enable === true) {
+                    node.serialConfig.emit('start');
+                } else {
+                    serialPool.close(node.serialConfig.serialport,() => {
+                        RED.log.info("[serialconfig:"+node.serialConfig.id+"] " + RED._("serial.stopped",{port:node.serialConfig.serialport}));
+                    });
+                }
+            }
+            node.send({payload: node.serialConfig});
```
dceejay, line 259: This is a specific serialConfig object so appears to debug as "[object Object]"... plus it has all the node internal config in it... This needs to be **just** the parameters you can change/set so that it could be simply modified and sent back into the control node. (as it was in your first iteration)

yhur:
Learning a lot 😉

dceejay:
Hi, There are a few small changes I want to make but rather than keep hassling you I will merge this as-is and then tweak before releasing. (Mainly I think the property should be named 'enabled' rather than 'enable', and no need to expose binary and char yet... If there is user demand they can be added - but right now they just add clutter.)

Thanks very much for your work on this

dceejay:
HI - I have pushed to master branch... I also added some status under the node
image
Will release later today (probably) after some more testing.

yhur:
@dceejay ,
Thank you very much.

